PR #5, opened 2019-11-13
Hotfix/4

davidhorak:
## Changes
- Invalid logical/comparison expression throw exception.

### Test Coverage
- [x] Unit tests


src/expression/comparison/gt.ts
```diff
@@ -28,6 +28,9 @@ export class GreaterThan extends Comparison {
    * @param {Operand} right Right operand.
    */
   constructor (left: Operand, right: Operand) {
+    if (arguments.length !== 2) {
```
johnpeb: future: We should consider adding a base class for binary operator where this can go

PR #36, opened 2020-06-18
Add Collection operand

### LINKED ISSUE #35: References not resolved when operand is an array
https://github.com/briza-insurance/illogical/blob/master/src/parser/options.ts#L73

This code seems incorrect when an operand is an array.

Example:
For the following condition references are not resolved
```
"condition": [
    "OVERLAP",
    [
      "$Location1ClassOfBusiness",
      "$Location2ClassOfBusiness",
      "$Location3ClassOfBusiness",
      "$Location4ClassOfBusiness",
      "$Location5ClassOfBusiness"
    ],
    [
      "561499-1-3",
      "561499-1-4"
    ]
  ]
```

davidhorak:
# Changes
- Add collection operand to resolve reference evaluation within array

## Test Coverage
- [x] unit

#### References
Closes #35

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
     name: Continuous integration
     runs-on: ubuntu-18.04
     container:
-      image: node:10.14.1-slim
+      image: node:12.14.1-slim
     steps:
       - uses: actions/checkout@v1
       - name: Install npm modules
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ npm-debug.log
 # VS Code
 *.vscode
 *.code-workspace
+
+# Webstorm
+.idea
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,8 @@
 # illogical changelog
 
+## 1.1.5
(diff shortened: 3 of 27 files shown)


lilyvanoekel:
In the tests for the overlap expression it's still like this: 
`{ left: new Value([1]), right: new Value([1, 2]), expected: true },`

Will this way stay supported? Would it be confusing to have both?

davidhorak:
> > In the tests for the overlap expression it's still like this:
> > `{ left: new Value([1]), right: new Value([1, 2]), expected: true },`
> > Will this way stay supported? Would it be confusing to have both?
> 
> Note, this should not be supported any more.
> As the **Collection** is handling array operands, so we may actually also update the **Value** to not accept array values anymore - as this is now a "never" case, due to this: https://github.com/briza-insurance/illogical/blob/feature/35/src/parser/index.ts#L303-L305
> 
> ... hence it should be:
> 
> ```
> { left: new Value([1]), right: new Collection([new Value(1), new Value(2)]), expected: true }
> ```

df51369fc5a9c0fa01f19f778024c619eaf00ddd

PR #39, opened 2020-06-22
Zero argument logical expressions

### LINKED ISSUE #38: Allow logical expression without arguments to be treated as array values
As showcased below, the expression throws an error as the "OR" keyword is treated as an expected operator. 

```json
["IN", "$state", ["OR"]]
```

Allow the logical expression without arguments, to be parsed as a collection.

davidhorak:
# Changes
- Allow zero argument logical expressions not be treated as collection.
- Allow logical expressions without any inner logical expressions to be treated as collection.

#### References
Closes #38

### DIFF
--- a/src/parser/index.ts
+++ b/src/parser/index.ts
@@ -145,10 +145,22 @@ export class Parser {
 
     /**
      * Simplify the logical expression if possible.
+     * - AND, OR with one operand is collapsed.
+     * - Logical expressions without operands are treated as collection.
+     * @param collapsible
      * @param operands
      */
-    const logicalExpressionReducer = (operands: Evaluable[]): Evaluable | undefined =>
-      operands.length === 1 ? operands[0] : undefined
+    const logicalExpressionReducer = (
+      collapsible: boolean,
+      operands: Evaluable[]
+    ): Evaluable | undefined => {
+      if (operands.length === 0) {
+        return this.getOperand(raw)
+      }
+      return collapsible && operands.length === 1
+        ? operands[0]
+        : undefined
+    }
 
     switch (operator) {
       /**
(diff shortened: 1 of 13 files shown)


src/parser/index.ts
```diff
@@ -145,31 +145,46 @@ export class Parser {
...
      * Simplify the logical expression if possible.
+     * - AND, OR with one operand is collapsed.
+     * - Logical expressions without operands are treated as collection.
+     * @param collapsible
      * @param operands
      */
-    const logicalExpressionReducer = (operands: Evaluable[]): Evaluable | undefined =>
-      operands.length === 1 ? operands[0] : undefined
+    const logicalExpressionReducer = (
+      collapsible: boolean,
```
rthreei, line 154: Why not make `collapsible` the second argument and optional with a default of `false`?

lilyvanoekel:
I guess if OR is the first state in a list of 3 that would still cause trouble right?

```
[
"OR",
"AL",
"AK"
]
```

But probably no way around that?

src/parser/index.ts
```diff
@@ -145,31 +145,46 @@ export class Parser {
...
+      operands: Evaluable[]
+    ): Evaluable | undefined => {
+      if (operands.length === 0) {
+        return this.getOperand(raw)
+      }
+      return collapsible && operands.length === 1
+        ? operands[0]
+        : undefined
+    }
```
rthreei, line 163: Does this account for this case:
```json
["IN", "$var", ["OR", "VT"]]
```

src/parser/index.ts
```diff
@@ -145,31 +145,46 @@ export class Parser {
 
     /**
      * Simplify the logical expression if possible.
+     * - AND, OR with one operand is collapsed.
+     * - Logical expressions without operands are treated as collection.
+     * @param collapsible
```
rthreei, line 150: Also a little lost by what `collapsible` is conveying.

davidhorak: 219b9f5a280fff7c8719b3b952ed3a352278f47a

src/parser/index.ts
davidhorak: 219b9f5a280fff7c8719b3b952ed3a352278f47a

src/parser/index.ts
davidhorak: It should now, a2f8e09c9ebee85ee86678853af369908227ae0b

davidhorak:
> I guess if OR is the first state in a list of 3 that would still cause trouble right?
> 
> ```
> [
> "OR",
> "AL",
> "AK"
> ]
> ```
> 
> But probably no way around that?

It should be good now, a2f8e09c9ebee85ee86678853af369908227ae0b ... you please double check ?

PR #41, opened 2020-07-22
Add present operator

### LINKED ISSUE #40: Support conditions that need to be agnostic of undefined/null
Currently we have the option to compare a value against `null`, or to use the `UNDEFINED` comparison expression.

It would be nice if we had a way to cover both cases easily. 

For example, the following condition:

```
"condition": [
    "NOR",
    [
      "UNDEFINED",
      "$Loss1Date"
    ],
    [
      "==",
      "$Loss1Date",
      null
    ]
  ]
```

Say it's possible for `Loss1Date` to be on the page but left empty, or maybe conditionally not shown at all, or filled out, and then emptied by the user. By the time it comes to evaluate the condition it might be either `null` or `undefined`. It would be nice if we can capture that in a single expression.

Some options:
* `UNDEFINED_OR_NULL` expression
* Updating `UNDEFINED` to check for null as well
* `EMPTY` / `NOT_EMPTY` expression
* `NIL` / `NOT_NIL` expression

lilyvanoekel:
Add present operator

Closes #40

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -46,6 +46,7 @@ yarn add @briza/illogical -D
     - [Suffix](#suffix)
     - [Overlap](#overlap)
     - [Undefined](#undefined)
+    - [Present](#present)
   - [Logical Expressions](#logical-expressions)
     - [And](#and)
     - [Or](#or)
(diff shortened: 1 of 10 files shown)


readme.md
```diff
@@ -387,6 +388,22 @@ engine.evaluate(['UNDEFINED', 'RefA'], { RefA: undefined }); // true
 engine.evaluate(['UNDEFINED', 'RefA'], { RefA: 10 }); // false
 ```
 
+#### Present
+Returns FALSE when the operand is UNDEFINED or NULL.
```
davidhorak, line 392: ~Returns~ Evaluates as FALSE ...

src/__test__/unit/index.test.ts
```diff
@@ -166,6 +166,33 @@ describe('Condition Engine', () => {
...
+        '$Name'
+      ],
+      inputs: [{
+        data: {
+          Name: undefined
+        },
+        expected: false
+      }, {
+        data: {
+          Name: 'David'
```
davidhorak, line 181: nice :D

readme.md
lilyvanoekel: 3eee647c81ce2d52620485e426f55e797ad7ca1d

rthreei:
This is nice 👍

PR #44, opened 2020-07-23
fix reference expression graceful handling of nils

andrewxu:
# Description
Currently it seems that the "in" (reference) expression would error if we are checking for isObject of a nil (null/undefined), this PR updates isObject to not throw error if the value is null or undefined, but instead, gracefully return false.

### Test Coverage
- [x] unit

#### References
- Closes https://github.com/briza-insurance/briza-api/issues/2774


src/common/__test__/unit/type-check.test.ts
```diff
@@ -42,4 +42,19 @@ describe('Condition Engine - Common - Type Check', () => {
...
+  test('isObject', () => {
+    let tests = [
+      { value: {}, expected: true },
+      { value: 'hi', expected: false },
+      { value: 1, expected: false },
+      { value: null, expected: false },
+      { value: undefined, expected: false }
+    ]
+
+    for (const test of tests) {
```
luiz290788, line 55: Suggestion: use [test.each](https://jestjs.io/docs/en/api#testeachtablename-fn-timeout)

src/common/type-check.ts
```diff
@@ -29,5 +29,9 @@ export function isString (value: Result): boolean {
  * @return {boolean} result of the test
  */
 export function isObject (value: unknown): boolean {
```
luiz290788, line 31: `export function isObject (value: unknown): value is Object {`

src/common/type-check.ts
```diff
@@ -29,5 +29,9 @@ export function isString (value: Result): boolean {
...
+  if (value === null || value === undefined) {
```
luiz290788, line 32: why not?
`return value !== null && value !== undefined && typeof value === 'object' && value.constructor === Object`

src/common/type-check.ts
andrewxu: is that proper TS for return type?
i'm getting:
type-check_ts______Documents_Development_illogical_src_common__-_VIM2

src/common/type-check.ts
andrewxu: typescript doesn't like `value.constructor` without checking for `value` as "value is possibly null",
and if we add "value" at the beginning of that chain, null && _ would always yield null.
also better readability, rule out the nil types and short-circuit

src/common/__test__/unit/type-check.test.ts
andrewxu: https://github.com/briza-insurance/illogical/pull/44/commits/5a4086712aa097403edcf190f9b46fb397edabda

src/common/type-check.ts
luiz290788: This is telling ts to assume that value is an `Object`. I've used object here but I should have used something else like the message is mentioning.

`export function isObject (value: unknown): value is Record<string, unknown> {`

src/common/type-check.ts
luiz290788: > rule out the nil types and short-circuit
In terms of short circuiting it is exactly the same.

You could do this:
```Typescript
  return value !== null && value !== undefined && value &&
      typeof value === 'object' && value.constructor === Object
```

But that's a bit weird for me because this could be simplified to what we had before:
```Typescript
return value && typeof value === 'object' && value.constructor === Object
```

src/common/__test__/unit/type-check.test.ts
```diff
@@ -1,45 +1,43 @@
...
-    ]
+  test.each([
+    ['1', true],
+    [new String('1'), true],
+    [1, false],
+    [true, false],
+    [false, false],
+    [{}, false],
+    [() => {}, false]
+  ])('isString(%p) should be %p.', (value, expected) => {
```
luiz290788, line 29: Loved it!

src/common/type-check.ts
andrewxu: yup that could work, and like you mentioned, it's pretty much going back to what we had before with the additional checks for the nils, i think the way it is currently written is slightly more readable because we explicitly call out the nil types and short circuit, rather than chaining 5 checks with &&.

If this really bothers you, i can change it.

src/common/type-check.ts
andrewxu: but why is the return value a record?

src/common/type-check.ts
luiz290788: I just don't think we need the explicit check for null and undefined when are checking for the boolean value of the object later on. That adds more complexity to the code.

If I eventually need to touch this file, I would remove that if, test would pass and I would commit that change. Just trying to understand the need for it.

src/common/type-check.ts
davidhorak: > export function isObject (value: unknown): value is Object

Yes, is is better, used in the API ... kinda legacy here, housekeeping to update to ```is X``` approach

> Record<string, unknown>

The newer version of TS has a rule for this one, to better use Record<string, unknown>``` rather than ```object```

luiz290788: The definition type of a record is
```Typescript
type Record<K, V> = {
  [K]: V
}
```

src/common/type-check.ts
andrewxu: https://github.com/briza-insurance/illogical/pull/44/commits/d2c47ec2a5e05f90281390250633c0c638347fc2

andrewxu:
> Do we need the check for null as well? Sorry for being picky.

yes we check both null and undefined.

PR #59, opened 2020-08-05
Support complex references

### LINKED ISSUE #57: Complex Reference Keys
## Description
We can explore a new illogical functionality to resolve complex reference keys to avoid more dynamic conditons.

#### Composite Reference Key
```$Q1{Q2}.state```
- Inner ```{}``` reference, is resolved first, therefore the final reference key would be: Q1**X**, where X is the evaluated reference, as string.

#### Nested references
```$Q1.{Q2}.state```
- any nested key, annotated with ```{}``` is evaluated first, and resolved as property string key.

#### Nested references via array index
```$Q1[{Q2}].state```
- any nested key, annotated with ```{}``` is evaluated first, and resolved as property array index.

### Concerns
- Prevent/detect cyclic references, e.g.: ```$Q1{Q1}```

### PR DESCRIPTION
**Pull request by lilyvanoekel**
# Description
* Support the complex reference keys described in #57 
* Readme update

### Test Coverage
- [x] unit

#### References
- Closes #57

### CODE REVIEW

**Comment by rthreei** [comment]
Cool feature.  Tidy implementation.

File: readme.md [hunk 1]
```diff
@@ -190,6 +190,17 @@ The evaluation data context is used to provide the expression with variable refe
 To reference the nested reference, please use "." delimiter, e.g.:
 ```$address.city```
 
+You can also use square brackets to access a specific element of an array:
```
davidhorak, line 193: Could we slightly reformulate here, if it make sense:

**Accessing Array Element:**
```$options[1]```

**Accessing Array Element via Reference:**
```$options[{index}]```
- **index** reference is resolved within the data context as an array index.

**Nested Referencing**
```$address.{segment}```
- **segment** reference is resolved within the data context as a property key.

**Composite Reference Key**
```$shape{shapeType}```
- **shapeType** reference is resolved within the data context, and inserted into the outer reference key.
- E.g. **shapeType** resolved as "**B**" would compose **$shapeB** outer reference.
- This resolution could be n-nested.

```
// Data context
const ctx = {
  name: 'peter',
  country: 'canada',
  age: 21,
  options: [1,2,3],
  address: {
    city: 'Toronto',
    country: 'Canada',
  },
  index: 2,
  segment: 'city',
  shapeA: 'box,
  shapeB: 'circle',
  shapeType: 'B'
}
```

File: readme.md [hunk 2]
```diff
@@ -198,17 +209,36 @@ const ctx = {
...
 engine.evaluate(['>', '$age', 20], ctx); // true
 
 // Evaluate an expression in the given data context
 engine.evaluate(['==', '$address.city', 'Toronto'], ctx); // true
+
+// Evaluate an expression using a reference that includes the array notation
+engine.evaluate(['==', '$options[1]', 2], ctx) // true
+
+// Evaluate an expression using a complex array reference
+engine.evaluate(['==', '$options[{optionIndex}]', 2], ctx) // true
```
andrewxu, line 235: since optionIndex is `[1]`, wouldn't this evaluate to `$options[[1]]`, and is that the desired?

lilyvanoekel: my bad, optionIndex should just be `1`.

Changing this as per David's comment

src/operand/reference.ts
```diff
@@ -16,6 +16,16 @@ import { Operand } from '.'
...
 function contextValueLookup (ctx: Context, key: string): Result {
+  // Resolve complex keys
+  const complexKeyExpression = /{([^{}]+)}/
+  let complexKeyMatches = complexKeyExpression.exec(key)
+
+  while (complexKeyMatches) {
+    const resolvedValue = contextValueLookup(ctx, complexKeyMatches[1])
+    key = key.replace(complexKeyExpression, `${resolvedValue}`)
+    complexKeyMatches = complexKeyExpression.exec(key)
+  }
```
davidhorak, line 27: thought: to make this lookup more effective, you really want to resolve the {} only for the given path segment (key[i]).
- e.g.: $a.{b}.{c}
- if the a would be resolved as undefined, you're returning right-away without the necessity to resolve the deeper references.

davidhorak: pretty much the same thing you're doing for arrays

src/operand/reference.ts
```diff
@@ -27,14 +37,29 @@ function contextValueLookup (ctx: Context, key: string): Result {
   let pointer = ctx
 
   for (let i = 0; i < keys.length; i++) {
+    let currentValue = pointer[keys[i]]
+
+    // Resolve array notation
+    const matchesArrayNotation = /^(.+)\[(\d+)]$/.exec(keys[i])
```
davidhorak, line 43: to be consistent, escape the "]" as well, i.e. ```\]```

src/operand/reference.ts
davidhorak: another small tip/though :)
the nested replacement could be simplified to:

```
key = key.replace(/{([^{}]+)}/, (match, innerReferenceKey) => contextValueLookup(ctx, innerReferenceKey))
```

readme.md
lilyvanoekel: 13b35e60462b79c6125c239e32eb7acd8406cdba

readme.md
lilyvanoekel: 13b35e60462b79c6125c239e32eb7acd8406cdba

src/operand/reference.ts
lilyvanoekel: 13b35e60462b79c6125c239e32eb7acd8406cdba

src/operand/reference.ts
lilyvanoekel: Would it be possible for there to be more than 1 'inner reference' per path segment?

davidhorak: It should be fine with "**g**" flag: https://regexr.com/59oa5 ... to sequentially resolve siblings pairs

lilyvanoekel: It fails on `Ref{Ref{RefE}}` though. And it doesn't handle the case where contextValueLookup returns undefined very well.

davidhorak: Why it fails on ```Ref{Ref{RefE}}```?

undefined:

```
try {
   key = key.replace(/{([^{}]+)}/, (match, innerReferenceKey) => {
      const resolvedKey = contextValueLookup(ctx, innerReferenceKey)
      if (!resolvedKey) { throw new Error(`failed to resolve inner reference: "${innerReferenceKey}"`) }
   })
} catch { return undefined }

... take my code example really just as inspiration

Comment (lilyvanoekel) [reply src/operand/reference.ts:27]: `Ref{Ref{RefE}}`

```
const ctx = {
  RefA: 'B',
  RefB: 'result',
  RefE: 'A',
  
}
```

If you only do one replace, it will only catch `{RefE}`. So the output will be `Ref{RefA}`.

If using a while loop, it will catch `{RefE}`, transform the key to `Ref{RefA}`, then on the next iteration it catches `{RefA}` and transforms the key to `RefB`

Also, even with the while loop. If I put it per path segment I'm having trouble with `RefG[{RefC.sub{RefD}}]`

Comment (davidhorak) [reply src/operand/reference.ts:27]: > `Ref{Ref{RefE}}`
> 
> ```
> const ctx = {
>   RefA: 'B',
>   RefB: 'result',
>   RefE: 'A'
> }
> ```
> 
> If you only do one replace, it will only catch `{RefE}`. So the output will be `Ref{RefA}`.
> 
> If using a while loop, it will catch `{RefE}`, transform the key to `Ref{RefA}`, then on the next iteration it catches `{RefA}` and transforms the key to `RefB`
> 
> Also, even with the while loop. If I put it per path segment I'm having trouble with `RefG[{RefC.sub{RefD}}]`
> 
> Edit:
> Also not a big fan of exceptions for flow control, unless it's in python :)

Make sense!

btw, here when **resolvedValue**  is undefined we should right-way return "undefined" no?
```
const resolvedValue = contextValueLookup(ctx, complexKeyMatches[1])
    key = key.replace(complexKeyExpression, `${resolvedValue}`)
```

Comment (lilyvanoekel) [reply src/operand/reference.ts:27]: > ```
> const resolvedValue = contextValueLookup(ctx, complexKeyMatches[1])
>     key = key.replace(complexKeyExpression, `${resolvedValue}`)
> ```

Yes, that make sense. I realised my original code was doing that `Refundefined` as a string thing too, hehe.

Comment (lilyvanoekel) [reply src/operand/reference.ts]: 468e5f97a097c75b2e6d4465929dd0934c591405

File: changelog.md [hunk 5]
```diff
@@ -1,5 +1,8 @@
 # illogical changelog
 
+## 1.2.4
+* Add support for complex references
```
davidhorak, line 4: As this is public package, let's be more verbose :)
* Add support for array element targeting withing reference operand key
* Add support for array element targeting via reference within reference operand key
* Add support for nested key resolution within reference operand operand key
* Add support for composite key resolution within reference operand operand key

or something like that ....

types/common/evaluable.d.ts
```diff
@@ -1,36 +1,36 @@
-/**
```
davidhorak, line 1: What is the different, tab vs spaces? I wonder if this is na PC vs MAC setup thing

lilyvanoekel: I think the files currently in the repository have windows line endings 
Screen Shot 2020-08-06 at 11 54 28 AM

davidhorak: hmm, will check on my end than

changelog.md
lilyvanoekel: 8a58c23a9ba9903ec1f55ead5d206252bc548aa1

src/operand/__test__/unit/reference.test.ts
```diff
@@ -30,7 +38,24 @@ describe('Operand - Value', () => {
...
+        ],
+        RefH: [
+          {
+            subA: 1
+          },
+          {
+            subA: 2
+          }
+        ],
+        Refundefined: 'A'
```
davidhorak, line 58: Let's remove this alien :)

lilyvanoekel: This was intentional though. My code was resolving the reference to `Refundefined`. It was returning undefined because `Refundefined` wasn't part of the context. So I added it, to make the test fail. Then I fixed the code to make the test pass again.

src/operand/reference.ts
```diff
@@ -27,14 +42,29 @@ function contextValueLookup (ctx: Context, key: string): Result {
   let pointer = ctx
 
   for (let i = 0; i < keys.length; i++) {
+    let currentValue = pointer[keys[i]]
+
+    // Resolve array notation
+    const matchesArrayNotation = /^(.+)\[(\d+)\]$/.exec(keys[i])
```
rajeevnaikte, line 48: Will it handle an array of an array? e.g. `RefH[0][1]`

lilyvanoekel: That's a good point, probably not. I guess this needs to be a loop as well

lilyvanoekel: 7f0434bc7fc616653450b2ed4d4885d306603d2d

lilyvanoekel: d4bb12e89809d55826168b69eb182bebef7ef84b

src/operand/reference.ts
```diff
@@ -27,14 +42,27 @@ function contextValueLookup (ctx: Context, key: string): Result {
   let pointer = ctx
 
   for (let i = 0; i < keys.length; i++) {
```
rajeevnaikte, line 44: Instead of this can we use lodash `get` method?

davidhorak: well, we're doing that resolution of inner references, etc. but potentially yes / maybe

src/operand/__test__/unit/reference.test.ts
lilyvanoekel: 03f6ece2d37ace1a91007c78367f9d73ae381548

src/operand/reference.ts
lilyvanoekel: Could work... do we want to add lodash as a dependency?

davidhorak: I would say, not keep the library as pure as possible, unless there is some signification benefit of using the lodash get

PR #137, opened 2021-02-10
Allow cross type parsing in Engine options

### LINKED ISSUE #135: Improve handling of numbers as strings
For comparison expressions like `>`, `<`, `>=`, etc. treat numbers encoded as strings as numbers, using `parseInt` or `parseFloat`.

We've done something similar for `wegood`:
https://github.com/briza-insurance/wegood/blob/master/src/rule/range.ts#L30-L41

theslyone:
# Description
Updates `eq`, `ge`, `gt`, `le`, `lt` and `ne` comparisons to allow cross type parsing if configured in Engine parser options

### Test Coverage
- [x] unit

#### References
- Closes #135


theslyone:
Ran the linter on the `.test.ts` files hence the reformatting, i wonder why linting test files were ignored.

rthreei:
The configuration is engine-wide, correct?  I suppose there's no elegant way to scope it to individual conditions.

theslyone:
> The configuration is engine-wide, correct? I suppose there's no elegant way to scope it to individual conditions.

Configuration is engine-wide yes but only the comparison conditions stated above check this flag

src/expression/comparison/eq.ts
```diff
@@ -18,18 +19,29 @@ export class Equal extends Comparison {
    * @param {Evaluable} left Left operand.
    * @param {Evaluable} right Right operand.
    */
-  constructor(...args: Evaluable[]);
-  constructor (left: Evaluable, right: Evaluable) {
-    if (arguments.length !== 2) {
-      throw new Error('comparison expression expects left and right operands')
+  private options: ComparisonOptions;
+  constructor(options: ComparisonOptions, ...args: Evaluable[]);
+  constructor (options: ComparisonOptions, left: Evaluable, right: Evaluable) {
```
davidhorak, line 24: not a big fan of changing the constructor and removing the arguments length check. Rather, I would offer this direction:

Add `strict` prop on the `Comparison` evaluable base, https://github.com/briza-insurance/illogical/blob/master/src/expression/comparison/index.ts ->

```ts
export abstract class Comparison implements Evaluable {
  type: EvaluableType = EvaluableType.Expression

  /**
   * Compared operands must be in the same data type to be successfully
   * evaluated and compared. E.g. [5, 5] vs [5, '5'].
   *
   * When set to false, the operands could some comparison operators will
   * try to cast types to the same type, e.g. [5, '5'] -> [5, 5] before the
   * evaluation. This may lead to unexpected behavior, should be used with caution.
   */
  public strict = true
```

by default set to true.

And change it to false, in the engine when operators are created as needed.

src/expression/comparison/index.ts
```diff
@@ -72,6 +83,24 @@ export abstract class Comparison implements Evaluable {
...
     }
-    return [operator, this.left.serialize(options), this.right.serialize(options)]
+    return [
+      operator,
+      this.left.serialize(options),
+      this.right.serialize(options)
+    ]
+  }
+
+  protected parseable (value: Result): number | undefined {
```
davidhorak, line 93: This more likely: `toNumber` and it should not return `undefined` rather the value it self in the default case. 

i.e. 

protected toNumber (value: Result): number | value {
   // ... parsing
   
   return value
}

davidhorak: Add a comment:

`Cast or convert a value to number if possible, otherwise return the input value untouched` or something like that

src/expression/comparison/eq.ts
```diff
@@ -18,18 +19,29 @@ export class Equal extends Comparison {
...
    */
   comparison (left: Result, right: Result): boolean {
+    if (this.options.allowCrossTypeParsing) {
+      left = this.parseable(left)
+      right = this.parseable(right)
+      if (isNumber(left) && isNumber(right)) {
+        return left === right
+      }
+    }
     return left === right
```
davidhorak, line 45: Considering these:
- https://github.com/briza-insurance/illogical/pull/137/files#r574117037
- https://github.com/briza-insurance/illogical/pull/137/files#r574120325

```ts

const operand = (value: Result) => this.strict ? value : this.toNumber(value)
return operand(left) === operand(right)

```

src/expression/comparison/index.ts
```diff
@@ -3,17 +3,26 @@
...
+  Result
+} from '../../common/evaluable'
+import { isEvaluable, isNumber, isString } from '../../common/type-check'
 import { Operand } from '../../operand'
 import { ExpressionInput } from '../../parser'
 import { Options } from '../../parser/options'
 
+export type ComparisonOptions = {
+  allowCrossTypeParsing: boolean;
+};
```
davidhorak, line 19: should not be needed, https://github.com/briza-insurance/illogical/pull/137/files#r574117037

src/parser/options.ts
```diff
@@ -67,6 +70,8 @@ export interface Options {
    */
   operatorMapping: Map<symbol, string>;
 
+  allowCrossTypeParsing: boolean;
```
davidhorak, line 73: as per this comment: https://github.com/briza-insurance/illogical/pull/137/files#r574117037

`strictTypeComparison: boolean`  ?

Add let's add documentation, this is a public lib.

davidhorak: And this should documented in the `readme.md` and you have to also regenerate the documentation `npm run docs`

src/expression/comparison/eq.ts
theslyone: 9b2e2c0

src/expression/comparison/index.ts
theslyone: 9b2e2c0

src/expression/comparison/index.ts
theslyone: 9b2e2c0

src/expression/comparison/eq.ts
theslyone: 9415a92

src/parser/options.ts
theslyone: 9b2e2c0, 6594c76

theslyone: `npm run docs` results in an error 

- `Error: Tried to set an option (mode) that was not declared.`
- `Error: Tried to set an option (target) that was not declared.`

There's a known bug with `typedoc` [here](https://github.com/TypeStrong/typedoc/issues/1447) and i noticed you bumped up from `0.20.19` to `0.20.10`. Will investigate further

theslyone: Release notes [here](https://github.com/TypeStrong/typedoc/releases/tag/v0.20.0) states 
> Removed options - mode, excludeNotExported, includeDeclarations, ignoreCompilerErrors (#1403), entryPoint

Resolution: Remove `mode` and `target` in the `typedoc.json` and regenerate docs d7fdee1

readme.md
```diff
@@ -660,6 +661,18 @@ import {
...
 
+#### Strict Type Comparison
+
+Configures strict data type check in comparison expressions
+
+```typescript
+strictTypeComparison: boolean;
+```
+
+**Default strict type comparison:**
```
davidhorak, line 672: Mention here what is the actual default value, to be clear. It is set to TRUE by default.

src/expression/comparison/index.ts
```diff
@@ -25,8 +30,10 @@ export abstract class Comparison implements Evaluable {
     protected readonly operator: string,
     protected readonly operatorSymbol: symbol,
     protected readonly left: Operand,
-    protected readonly right: Operand) {
-  }
+    protected readonly right: Operand
+  ) {}
+
+  public strict = true;
```
davidhorak, line 36: Please add the comment, as suggested (you can use your words). You should really describe the meaning, and default behavior. (as open source, we should be really clear)

src/expression/comparison/index.ts
```diff
@@ -72,6 +81,28 @@ export abstract class Comparison implements Evaluable {
...
+    if (isValueNumber) {
+      return value as number
+    } else if (isString(value)) {
+      if (value.match(/^\d+\.\d+$/)) {
+        return parseFloat(value)
+      } else if (value.match(/^0$|^[1-9]\d*$/)) {
+        return parseInt(value)
+      }
+    }
+    return value
```
davidhorak, line 106: As this function does not depend on state of the class, we should separate it ... just have as a utility function.

src/expression/comparison/__test__/unit/eq.test.ts
```diff
@@ -1,30 +1,37 @@
-import { Result } from '../../../../common/evaluable'
+import {
```
lilyvanoekel, line 1: Shouldn't strict/not strict comparison cases both be tested for `eq` as well?

src/expression/comparison/ge.ts
```diff
@@ -31,9 +31,13 @@ export class GreaterThanOrEqual extends Comparison {
    * {@link Comparison.comparison}
    */
   comparison (left: Result, right: Result): boolean {
-    if (isNumber(left) && isNumber(right)) {
+    const operand = (value: Result) =>
+      this.strict ? value : this.toNumber(value)
+
+    if (isNumber(operand(left)) && isNumber(operand(right))) {
       return (left as number) >= (right as number)
```
davidhorak, line 38: This will work as expected - unless I am missing something.
`isNumber(operand(left))`
 - if typed and evaluated as number, good 
 - but than here `(left as number)` you are still using the original variable - i.e. not typed to number,

src/expression/comparison/gt.ts
```diff
@@ -31,9 +31,13 @@ export class GreaterThan extends Comparison {
    * {@link Comparison.comparison}
    */
   comparison (left: Result, right: Result): boolean {
-    if (isNumber(left) && isNumber(right)) {
+    const operand = (value: Result) =>
+      this.strict ? value : this.toNumber(value)
+
+    if (isNumber(operand(left)) && isNumber(operand(right))) {
```
davidhorak, line 37: ditto, https://github.com/briza-insurance/illogical/pull/137/files#r574506004

src/expression/comparison/le.ts
```diff
@@ -31,9 +31,13 @@ export class LessThanOrEqual extends Comparison {
    * {@link Comparison.comparison}
    */
   comparison (left: Result, right: Result): boolean {
-    if (isNumber(left) && isNumber(right)) {
+    const operand = (value: Result) =>
+      this.strict ? value : this.toNumber(value)
+
+    if (isNumber(operand(left)) && isNumber(operand(right))) {
```
davidhorak, line 37: ditto, https://github.com/briza-insurance/illogical/pull/137/files#r574506004

src/expression/comparison/lt.ts
```diff
@@ -31,9 +31,13 @@ export class LessThan extends Comparison {
    * {@link Comparison.comparison}
    */
   comparison (left: Result, right: Result): boolean {
-    if (isNumber(left) && isNumber(right)) {
+    const operand = (value: Result) =>
+      this.strict ? value : this.toNumber(value)
+
+    if (isNumber(operand(left)) && isNumber(operand(right))) {
```
davidhorak, line 37: ditto, https://github.com/briza-insurance/illogical/pull/137/files#r574506004

src/parser/index.ts
```diff
@@ -114,14 +151,26 @@ export class Parser {
...
+     * Creates an T instance of a comparison expression and ensures that strict option is configured
+     */
+    const getComparisonExpression = <T extends Comparison>(
+      Expression: new (...o: Evaluable[]) => T,
+      operands: Evaluable[]
+    ) => {
+      const expression = new Expression(...operands)
+      expression.strict = this.opts.strictTypeComparison
+      return expression
     }
```
davidhorak, line 174: 💪🏻

src/expression/comparison/__test__/unit/eq.test.ts
davidhorak: 💯 ... And this https://github.com/briza-insurance/illogical/pull/137#discussion_r574506004 I guess be captured.

src/expression/comparison/ge.ts
lilyvanoekel: I guess it "works as expected" because it's tricking typescript. For regular javascript `2 > '1'` works, so the tests pass, even if `toNumber` is not used for the actual comparison.

lilyvanoekel:
Should we have test cases for date comparison `'2010-01-01' > '2009-02-03'`?

src/expression/comparison/ge.ts
davidhorak: TIL: `'2.3' > '2.2'` is valid :D

davidhorak:
> Should we have test cases for date comparison `'2010-01-01' > '2009-02-03'`?

Part of this PR?

lilyvanoekel:
> Part of this PR?

I would expect that to work even with strict mode off, comparing 2 strings.

src/expression/comparison/ge.ts
theslyone: > TIL: `'2.3' > '2.2'` is valid :D

I was also shocked to see that this works, '2010-01-01' < '2009-02-03' as well

theslyone:
> Should we have test cases for date comparison `'2010-01-01' > '2009-02-03'`?

This works with or without the strict mode, i can add this case to the test to verify

src/expression/comparison/ge.ts
davidhorak: Okay, so I would actually propose that it in `strict` mode (default):
 - `ge, gt, le, lt` would evaluate to false, if any of the operands is **NOT** a numbers
 - i.e. the same behavior as `eq, ne` (achieved by the === value and type check natively by JS)
- This is avoid these unexpected "side effects" from the DSL perspective.

@RyanNieuwoudt @lilyvanoekel @theslyone thoughts?

theslyone: This will be a breaking change because default mode is strict but will cause `'2010-01-01' > '2009-02-03' = false` and `2 > '1' = false` when js by default allows this to pass

src/expression/comparison/index.ts
theslyone: 46e890e

src/expression/comparison/ge.ts
davidhorak: Understand, but `JS` implementation of the illogical DSL. If 3rd party will implement our questionnaire it other languages, that would not expect `2 > '1'` to evaluate as true.

And, do we actually some conditions which will break?

theslyone: > Understand, but this is just JS implementation of the illogical DSL. If 3rd party will implement our questionnaire in other languages, than they would not expect 2 > '1' to evaluate as true.

Do you know of any third party usage of this particular js library currently? I'm saying this will break their implementation and so we need to capture this in our breaking changes.

> And, do we actually some conditions which will break?

We cannot be certain at this point the implication of this on briza-api although we plan to use it with `strict=false` so we should be good

lilyvanoekel: I think without @theslyone changes it checks if things are a number so `'2010-01-01' > '2009-02-03'` will return false?

I'm not so sure about other languages and what happens to `'2010-01-01' > '2009-02-03'`. This seems like a normal string compare, no?

Agree on `2 > '1'`

lilyvanoekel: https://stackoverflow.com/questions/8406703/c-compare-to-string-dates

davidhorak: > We cannot be certain at this point the implication of this on briza-api although we plan to use it with strict=false so we should be good

This should be testable. "Simply" traversing trough all questions condition props - @RyanNieuwoudt you've just written something like this for detection of conflicts in conditions - any suggestions?

davidhorak: > Do you know of any third party usage of this particular js library currently? I'm saying this will break their implementation and so we need to capture this in our breaking changes.

Not sure, will double check. Though, it would "break" only if we have some issues in the conditions which is TBD, and we can fix them when detected.

lilyvanoekel: maybe we should not have strict mode. Instead just force both operands to be of the same type?

We should probably also define what we want to happen with `null` and `undefined`

davidhorak: > '2010-01-01' > '2009-02-03'. This seems like a normal string compare, no? Yes, string compare (though even that is not normal in other major languages) but it evokes the feeling that it is actually comparing dates.

theslyone: > maybe we should not have strict mode. Instead just force both operands to be of the same type?

This will bring us back to our original issues below wherein we'll be relying solely on js to ensure `'10' > '0' = true` 
```ts
[
  ">",
  "$Location1BuildingLimit",
  "0"
]```

lilyvanoekel: We could specifically check for the types number or string. I'm sure every language has some kind of string compare function.

readme.md
theslyone: d448793

src/expression/comparison/index.ts
theslyone: d448793

src/expression/comparison/ge.ts
davidhorak: > [">", "$Location1BuildingLimit", "0"]

we should fix these to `[">", "$Location1BuildingLimit", 0]`

theslyone: > we should fix these to [">", "$Location1BuildingLimit", 0]

`$Location1BuildingLimit` is a `currency` question type so answer is a string hence the problem.

We want to enable type coercion by setting `strict = false` so that illogical can resolve `[">", "$Location1BuildingLimit", 0]` properly

davidhorak: > Do you know of any third party usage of this particular js library currently?

Quick intel from Luiz (not to be taken word by word): "TowerIQ is java and I believe that they were going to implement their own version of the condition evaluator (Rishi might have better details). Quiktract is php and I also believe that they were going to implement their own version. I can check with Nick"

davidhorak: > $Location1BuildingLimit is a currency question type so answer is a string hence the problem.

I see. 

Shall we than define/describe that `ge, gt, le, lt` should convert parse-able strings to numbers?
- By definition of the illogical DSL?
- Therefore actually remove the `strict` mode, and rather follow this set standard.
- If 3rd parties will implement illogical DSL, than they would understand what is expected.

davidhorak: Another very interesting idea (@undees): 

Explicit optional reference casting: `[">", "$Ref.(Number)", 0]`

theslyone: I'll investigate this and probably close this PR in favour of another to avoid :wrench: :hammer: :fire: :fire:

davidhorak: Btw, it may worthwhile to create first a PR which will include only the prettier formatting update - to make the PR super focused on the actual develop feature.

RyanNieuwoudt: I really like where @undees is going with that. Could this include the ability to perform any `map` operation on the data before the comparison takes place?

Allowing for explicit conversions:

1. Guards against the principle of least surprise (for non JavaScript developers in particular).
2. Could be more flexible than just introducing JavaScript's coercion behaviours.
3. May require less code in Illogical itself, deferring to the consuming code for circumstantial logic.

davidhorak: > Avoids adding the strictMode option

In this case https://github.com/briza-insurance/illogical/pull/137#discussion_r574627064 ... it should not be needed.

theslyone:
Closing in favor of https://github.com/briza-insurance/illogical/pull/139

PR #139, opened 2021-02-12
Allow $Ref casting

### LINKED ISSUE #135: Improve handling of numbers as strings

theslyone:
# Description
- Add support for $ref casting

### Test Coverage
- [x] unit

#### References
- Closes #135


src/operand/reference.ts
```diff
@@ -102,6 +119,13 @@ export class Reference extends Operand {
     }
     super()
     this.key = key
+
+    const dataTypeRegex = /^.+\.\((Number|String|Date)\)$/
```
theslyone, line 123: Hopefully we can support date comparison at some point

src/operand/reference.ts
```diff
@@ -86,11 +87,27 @@ function contextValueLookup(ctx: Context, key: string): Result {
...
+ * @param value value to parse as data type
+ */
+function toDataType(value: Result, dataType: string | undefined): Result {
+  switch (dataType) {
+    case 'Number':
+      return toNumber(value)
+    case 'String':
+      return toString(value)
+    default:
+      return value
```
theslyone, line 101: Should we consider throwing an error here?

src/operand/__test__/unit/reference.test.ts
```diff
@@ -66,6 +67,11 @@ describe('Operand - Value', () => {
       ['RefA{RefA}', undefined],
       ['RefB.{RefA}', undefined],
       ['Ref{RefB}', undefined],
+      // Data type parsing
+      ['RefH[{RefA}].sub{RefD}.(Number)', 2],
+      ['RefA.(String)', '1'],
+      ['RefJ.(String)', '1'],
+      ['RefJ.(Number)', 1],
```
rthreei, line 74: For familiarity, go the way of Typescript instead? e.g. `<Number>RefJ`.  It's also one less character (not that it really matters).

rthreei (commented):
Love the new approach.  Smaller in scope and targeted.

src/operand/__test__/unit/reference.test.ts
theslyone: Idea is for illogical's DSL to be language agnostic considering that there will probably be implementations in other languages as well.

rthreei: Sure, but not sure if I've seen casting done as `.(type)` in languages.

src/operand/reference.ts
lilyvanoekel: If it's for "at some point" maybe we shouldn't include `Date` now?

theslyone: sure, just a POC, can eventually have this implemented in this PR if agreed

davidhorak: I agree, the `Date` is a bit more complex as we have to also define which format of date would be supported for this casting.

src/operand/__test__/unit/reference.test.ts
lilyvanoekel: http://rigaux.org/language-study/syntax-across-languages.html#TpsCast

`(Type) $ref`?

src/operand/reference.ts
davidhorak: Maybe not throwing but console.warn would be nice! To give us the chance to fix these

src/operand/__test__/unit/reference.test.ts
davidhorak: :) https://golang.org/ref/spec#Type_assertions

davidhorak: > Idea is for illogical's DSL to be language agnostic considering that there will probably be implementations in other languages as well.

💯 ... though really the format does not matter, if easy to understand / implement

src/operand/reference.ts
```diff
@@ -86,11 +87,27 @@ function contextValueLookup(ctx: Context, key: string): Result {
...
+/**
+ * Converts a value to a specified data type, returns original value if not parseable.
+ * @param value value to parse as data type
+ */
+function toDataType(value: Result, dataType: string | undefined): Result {
+  switch (dataType) {
+    case 'Number':
+      return toNumber(value)
+    case 'String':
+      return toString(value)
```
davidhorak, line 99: I guess, also for Number and String, we may console.warn as well, if evaluated as `undefined` - really to give us the change to fix these, as we define the conditions.

src/operand/__test__/unit/reference.test.ts
rthreei: I stand corrected!  I'll leave it to you folks to duel it out.

davidhorak: @theslyone you can create a quick pool in the #carriers-channel to get votes :D

theslyone: > @theslyone you can create a quick pool in the #carriers-channel to get votes :D

Hopefully will wouldn't have to drag this decision to slack, i personally think `$ref.(Type)`, `$ref.<Type>`, `<Type>.$ref`, `(Type)$ref`, `<Number>$ref` and all it's variants work fine and we should just stick to one and have the readme/docs updated appropriately.

davidhorak: Yeah. I have to strong preference for any of those, so I will it keep it up to your decision.

src/operand/reference.ts
theslyone: b8f7162

src/operand/reference.ts
theslyone: b8f7162

src/operand/reference.ts
theslyone: 58aaeca

theslyone:
Also worth pulling up again, @RyanNieuwoudt had brought up an interesting idea [here](https://github.com/briza-insurance/illogical/pull/137#discussion_r574653778)
>I had been wondering whether the problem could be solved by means of 'transform hooks', i.e. a mechanism in Illogical that would call user code before the expression is evaluated and allow for arbitrary transformations (type casting, string cleanup, whatever) to take place first.

This will allow for evaluation/cleanup of the operands pre evaluation by illogical at runtime

### Example usercases

#### Complex evaluations

```js
[
  ">",
  "YEAR($BusinessStartDate)",
  "MINUS(CURRENT_YEAR(), 3)"
]
```
instead of having this resolved by illogical (which indeed it shouldn't), we can implement hooks into the operand resolver, pass in the operand as a parameter (in above case return new operand for `"YEAR($BusinessStartDate)"` as year component of `$BusinessStartDate`, new operand for `"MINUS(CURRENT_YEAR(), 3)"` as currentYear - 3).

#### Date comparison

```
[
  ">",
  "$BusinessStartDate",
  "2009-02-03"
]
```
can also easily be handled by first converting both operands to unix timestamps.

I feel having such flexibility will allow illogical's `comparison` expression evaluators `ge`, `gt` etc work as expected by delegating the operand cleanups to the calling code.

Hope i'm speaking English and not Korean? :laughing:

davidhorak:
> Also worth pulling up again, @RyanNieuwoudt had brought up an interesting idea [here](https://github.com/briza-insurance/illogical/pull/137#discussion_r574653778)
> 
> > I had been wondering whether the problem could be solved by means of 'transform hooks', i.e. a mechanism in Illogical that would call user code before the expression is evaluated and allow for arbitrary transformations (type casting, string cleanup, whatever) to take place first.
> 
> This will allow for evaluation/cleanup of the operands pre evaluation by illogical at runtime
> 
> ### Example usercases
> #### Complex evaluations
> ```js
> [
>   ">",
>   "YEAR($BusinessStartDate)",
>   "MINUS(CURRENT_YEAR(), 3)"
> ]
> ```
> 
> instead of having this resolved by illogical (which indeed it shouldn't), we can implement hooks into the operand resolver, pass in the operand as a parameter (in above case return new operand for `"YEAR($BusinessStartDate)"` as year component of `$BusinessStartDate`, new operand for `"MINUS(CURRENT_YEAR(), 3)"` as currentYear - 3).
> 
> #### Date comparison
> ```
> [
>   ">",
>   "$BusinessStartDate",
>   "2009-02-03"
> ]
> ```
> 
> can also easily be handled by first converting both operands to unix timestamps.
> 
> I feel having such flexibility will allow illogical's `comparison` expression evaluators `ge`, `gt` etc work as expected by delegating the operand cleanups to the calling code.
> 
> Hope i'm speaking English and not Korean? 😆

This one is very interesting, so pretty much dynamic registration of function expressions/evaluation - worth of separate discussion.
- though, one quick note, doing so, this library would not anymore would be able to evaluate correctly the briza questionnaire by itself.

src/common/util.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+  return undefined
+}
+
+/**
+ * Convert a value to string if possible, otherwise return undefined
+ * @param value value to be converted to string
+ */
+export const toString = (value: Result): string | undefined => {
+  return value ? `${value}` : undefined
```
lilyvanoekel, line 27: should an empty string return `undefined`?

readme.md
```diff
@@ -240,6 +240,12 @@ To reference the nested reference, please use "." delimiter, e.g.:
 * E.g. **shapeType** is resolved as "**B**" and would compose the **$shapeB** outer reference.
 * This resolution could be n-nested.
 
+#### Data Type Casting
+`$payment.amount.(Number)`
+
+-    This forces the data contexts `payment.amount` to be cast into a `number` before being used as an operand in the evaluation.
+-     Support data type cas ``(Number)`` <=> ``(String)``
```
davidhorak, line 247: Suggestions: Just a small updates for clarity

`$payment.amount.(Type)`

Cast the given data context into the desired data type before being used as an operand in the evaluation.
> Note: If the conversion is invalid, the an warning message is being logged.

Supported data type conversions: 
- `.(String)`: cast a given reference to **String**.
- `.(Number)`: cast a given reference to **Number**.

src/common/util.ts
davidhorak: I would say no, an empty string is a valid string

src/operand/reference.ts
```diff
@@ -86,11 +87,44 @@ function contextValueLookup(ctx: Context, key: string): Result {
...
 
+enum DataType {
+  Number = 'Number',
+  String = 'String',
+}
+
+// Equivalent to /^.+\.\((Number|String)\)$/
+const dataTypeRegex = new RegExp(
+  `^.+\\.\\((${Object.keys(DataType).join('|')})\\)$`
+)
```
davidhorak, line 98: Minor (not action at this moment): I would wonder if we could convert this into a non-rx solution to save some performance.

src/operand/reference.ts
```diff
@@ -86,11 +87,44 @@ function contextValueLookup(ctx: Context, key: string): Result {
...
+    case DataType.Number:
+      result = toNumber(value)
+      break
+    case DataType.String:
+      result = toString(value)
+      break
+  }
+  if (value && dataType && result === undefined) {
+    console.warn(`Casting ${value} to ${dataType} resulted in ${result}`)
+  }
```
davidhorak, line 118: Could we please unit test this?

src/common/util.ts
lilyvanoekel: I worded that too carefully, I believe that will return `undefined` for empty strings.

theslyone: I question why someone will need to cast a string to a string by using `.(String)` in the condition. Expectation is that the cast usage is solely for conversion from one type to another. I've update to support `0` though.

cc8306c

davidhorak: It may simply a mistake, or invalid data context, so better to capture it.

readme.md
theslyone: 7185af3

src/common/util.ts
lilyvanoekel: Definitely agree on `0 => '0'`, but I don't think we should reason our way out of fixing empty strings. Sometimes it's hard to reason about the path the code will take (today or in the future). I would not expect `toString` to return `undefined` for empty strings, and it could be a potential future source of bugs. However unlikely, I think it should be fixed. Maybe make the check `is nil` for null and undefined specifically instead?

lilyvanoekel (approved):
LGTM provided David's feedback is addressed

src/operand/reference.ts
theslyone: e1166d2

davidhorak:
> had a discussion with Lily about it, should we then result to returning undefined here?

Yes, simply not coverable, **at least at this moment - in this PR**, to number.

theslyone:
> Yes, simply not coverable, **at least at this moment - in this PR**, to number.

I've updated. This PR kinda blocks a lot of LM questions so hopefully i can get this over with today

davidhorak (approved):

davidhorak:
LGTM!

When released, let's double check that API and Binder still works as expected

undees:
🙌🏻 Nice work! Thank you for working through all these permutations.

PR #183, opened 2021-05-17
set babel env targets

luiz290788:
# Description

Setting the target of the `@babel/env` preset to `> 1%, node 12`. This means browsers with trafic greater than 1% and node 12.

Why?

The default target is to transpile everything to support ES5 and this was causing performance issues in projects using illogical in node.

### Test Coverage
- [x] unit

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,8 @@
 {
   "presets": [
-    "@babel/env",
+    ["@babel/env", {
+      "targets": "> 1%, node 12"
+    }],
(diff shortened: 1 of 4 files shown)


luiz290788:
I guess I will have to update dependencies first.

davidhorak (approved):
Looks reasonable to me.

We may add into the readme, the information about used `targets`, and note that the library could be re-build with different `targets`.

davidhorak (approved):
🚀

PR #194, opened 2021-06-17
Simplification of codebase, added some extra features

davidhorak:
## Description
Simplification of codebase, added some extra features (a weekend project, to do some coding, intention is to just share this with team, proceed with the PR only if found valuable)

**Notes:**
- The base functionality is preserved
- Better to review locally (due to large volume of file changes)

### Changes
- Removed classes, migrated to pure functions
   - The output library size is reduced from `42kb` to `24kb`
- Unlimited expression nesting, i.e. comparison expression operands now could be arbitrary expressions
  - e.g.: `[==, $RefA, [==, 1, 1]]`
- Any primitive values could be now evaluated, i.e. the parser does not restrict the expression input to be a logical expression
  - e.g.: `illogical.evaluate(1)` evaluated as `1`
- Collections starting with logical or comparison expression operator symbol could be now escaped to be processed as regular collections
  - e.g.: `[IN, $RefA, [\==, 1, 1]]` ... `\` used as escape character (configurable)
- Small refactoring of options
- Simplification of references now allows to use regular expressions in `ignoredPaths`
- Complete test coverage

### Test Coverage
- [x] unit

Screenshot 2021-06-17 112139

### Potential TODO
- [ ] Update documentation
- [ ] Regenerate the types

PR #210, opened 2021-08-17
Change simplify to allow complementary defer evaluate behavior

rpedrosanto:
# Description

The purpose of this change is to suggest a better name of `strictKeys` to the existing `ignoreKeys` parameter while also extending the feature of `Evaluable.simplify` by adding a new `optionalKeys` parameter which provides a complementary behavior to the first allowing to specify keys whose evaluation should be postponed while everything else should be considered present.

~~Note: typedoc is not being updated in this PR due to the issue described in https://github.com/briza-insurance/illogical/issues/211~~ Docs updated

### Test Coverage
- [x] unit

#### References
- N/A


src/common/evaluable.ts
```diff
@@ -60,10 +60,17 @@ export interface Evaluable {
...
+   * @param {false | string[]} deferEvaluate keys to be considered not present unless they are also in `alwaysEvaluate`;
+   *  when `alwaysEvaluate` is set to `false` and `deferEvaluate` is an array, every key that is not in `deferEvaluate`
+   *  is considered to be present and thus will be evaluated
    * @returns {Result | Evaluable} simplified value or itself
    */
-  simplify(ctx: Context, ignoreKeys: string[]): Result | Evaluable
+  simplify(
+    ctx: Context,
+    alwaysEvaluate?: false | string[],
+    deferEvaluate?: false | string[]
```
rthreei, line 72: Thinking about naming.. would `strictKeys?: string[]` and `skipKeys?: string[]` be better?

davidhorak, line 72: For consideration, https://github.com/briza-insurance/illogical/blob/e7405e310742e179ef0d4981669b4b4a55064e0b/src/operand/reference.ts#L30-L32 to extend here to support **regexp** (small bit form that experimental PR: https://github.com/briza-insurance/illogical/pull/194)

davidhorak: +1 `strict | optional`

luiz290788: Interesting. I think it would be backward compatible if in the future we add RegExp.

davidhorak: Yeah, it should. Just for consideration here.

rpedrosanto: Done a292b90d26f2c82ea76701a5aacdc81f7e2a6674

src/common/evaluable.ts
```diff
@@ -60,16 +60,16 @@ export interface Evaluable {
...
+   *  when `strictKeys` is set to `false` and `optionalKeys` is an array, every key that is not in `optionalKeys`
    *  is considered to be present and thus will be evaluated
    * @returns {Result | Evaluable} simplified value or itself
    */
   simplify(
     ctx: Context,
-    alwaysEvaluate?: false | string[],
-    deferEvaluate?: false | string[]
+    strictKeys?: false | string[],
+    optionalKeys?: false | string[]
```
rthreei, line 72: Probably should be reversed?

`extraKeys?: string[]` - new name for `ignoreKeys`

`strictKeys?: string[]`

Also, is `false |` necessary?  Given that the argument is optional?  So you can test whether it is `undefined` instead of `false`.

src/operand/reference.ts
```diff
@@ -137,15 +137,22 @@ export class Reference extends Operand {
...
+  ): Result | Evaluable {
     const keys = extractKeys(ctx, this.key)
 
     if (!keys) {
       return this
     }
 
     const key = keys[0].replace(/\[.+$/, '')
-    if (ctx[key] !== undefined || ignoreKeys.includes(key)) {
+    const doEval =
```
rthreei, line 152: `shouldEvaluate`?

changelog.md
```diff
@@ -1,5 +1,10 @@
 # illogical changelog
 
+## 1.5.0
+
+- Add simplify capability to evaluate the expression for all referred values
+  not present in the context except for a specified list of keys
```
davidhorak, line 6: We may tweak this as well to be aligned with the new thermology (strict, optional).

readme.md
```diff
@@ -201,10 +201,24 @@ in the context.
 engine.simplify(
   ['AND', ['==', '$a', 10], ['==', '$b', 20]],
   { a: 10 },
-  ['b'] // '$b' will be evaluted to undefined.
+  ['b'] // '$b' will be evaluated to undefined.
 ) // false
 ```
 
+Alternatively we might want to do the opposite and evaluate the expression for all referred values
```
davidhorak, line 208: ditto, https://github.com/briza-insurance/illogical/pull/210/files#r691178782

src/common/evaluable.ts
rpedrosanto: I don't think I get it. In my opinion `strictKeys` instead os `ignoreKey` makes more sense as those are the keys that should be strictly evaluated no matter if they are in the context or not.

As for `optionalKeys` are those that could be evaluated if they are in the context or a part of the `strictKeys`.

What's in your mind?

src/common/evaluable.ts
```diff
@@ -60,10 +60,17 @@ export interface Evaluable {
...
    * @param {Context} ctx context for the evaluation
-   * @param {string[]} ignoreKeys keys to be considered present even if their not present in the context
+   * @param {false | string[]} strictKeys keys to be considered present even if they are not present in the context
+   * @param {false | string[]} optionalKeys keys to be considered not present unless they are also in `strictKeys`;
+   *  when `strictKeys` is set to `false` and `optionalKeys` is an array, every key that is not in `optionalKeys`
+   *  is considered to be present and thus will be evaluated
    * @returns {Result | Evaluable} simplified value or itself
    */
-  simplify(ctx: Context, ignoreKeys: string[]): Result | Evaluable
+  simplify(
```
davidhorak, line 69: Another small thing for consideration: https://github.com/briza-insurance/illogical/blob/e7405e310742e179ef0d4981669b4b4a55064e0b/src/evaluable.ts#L48-L49
- Should we have simplification / serialization options (object), rather that expanding the args?
- As we may be adding features for simplification/serialization it maybe easier to just expand the options object param.

src/operand/reference.ts
```diff
@@ -137,15 +137,22 @@ export class Reference extends Operand {
...
+      (strictKeys !== false && strictKeys.includes(key)) ||
+      (optionalKeys !== false && !optionalKeys.includes(key))
```
davidhorak, line 154: Would ```(strictKeys && strictKeys.includes(key)) || (optionalKeys && !optionalKeys.includes(key))``` be sufficient here?

src/common/evaluable.ts
rthreei: > I don't think I get it. In my opinion `strictKeys` instead os `ignoreKey` makes more sense as those are the keys that should be strictly evaluated no matter if they are in the context or not.
> 
> As for `optionalKeys` are those that could be evaluated if they are in the context or a part of the `strictKeys`.
> 
> What's in your mind?

Ah, yeah you're right.  That makes sense.  Only thing then:

> Also, is false | necessary? Given that the argument is optional? So you can test whether it is undefined instead of false.

rpedrosanto: It's not, I'll remove the `false` and follow David's recommendation [here](https://github.com/briza-insurance/illogical/pull/210#discussion_r691188204)

src/common/evaluable.ts
luiz290788: I like it be it would be a breaking change if we introduce it now.

To avoid the breaking change. We would need to do something like:

```
simplify(context: Context, optionsOrIgnoreKeys: string[] | SimplifyOptions): Evaluated | Evaluable
```

Also hiding this in options might be hard for the developer to realise he is introducing a breaking change. Changing a type of a property in the options will be a breaking change but is hard to see that.

I would not do it as part of this PR.

I believe that illogical needs a little bit of a revamp. There is the idea of getting rid of the classes and make it fully functional. We might want to push these kind of change to a newer major version of illogical. We could create this newer version without changing the evaluable specification, just change the implementation.

src/operand/reference.ts
```diff
@@ -137,15 +137,22 @@ export class Reference extends Operand {
...
     }
 
     const key = keys[0].replace(/\[.+$/, '')
-    if (ctx[key] !== undefined || ignoreKeys.includes(key)) {
+    const shouldEvaluate =
+      (strictKeys && strictKeys.includes(key)) ||
+      (optionalKeys && !optionalKeys.includes(key))
+    if (ctx[key] !== undefined || shouldEvaluate) {
       return this.evaluate(ctx)
     }
```
luiz290788, line 157: A small improvement:
- we don't need to call `this.evaluate` if `shouldEvaluate` is false. We know that it will return undefined. There is the case of complex keys, but that is not being handled here anyway.
- We don't need to check if the key is included when we know that the value is defined.

```suggestion
    if (ctx[key] !== undefined || shouldEvaluate) {
      return this.evaluate(ctx)
    } else if (
      (strictKeys && strictKeys.includes(key)) ||
      (optionalKeys && !optionalKeys.includes(key))
    ) {
      return undefined
    }
```

rpedrosanto: Makes sense. Done in 5146b749ff24c2b6ac3efc2e538e678f39fbefcd

src/common/evaluable.ts
```diff
@@ -55,10 +55,17 @@ export interface Evaluable {
    * Simplifies this Evaluable when possible.
    *
    * @param {Context} ctx context for the evaluation
-   * @param {string[]} ignoreKeys keys to be considered present even if their not present in the context
+   * @param {string[]} strictKeys keys to be considered present even if they are not present in the context
+   * @param {string[]} optionalKeys keys to be considered not present unless they are also in `strictKeys`;
+   *  when `strictKeys` is `undefined` and `optionalKeys` is an array, every key that is not in `optionalKeys`
+   *  is considered to be present and thus will be evaluated
```
luiz290788, line 61: Reading this description carefully, it doesn't seem to reflect exact what is happening. Because we only check `optionalKeys` when the value is not present on the context.

rpedrosanto: Done 245d7e99a7819e75c4d48d14697fc2d608bd5c51

luiz290788 (approved):
LGTM

@davidhorak and @rthreei can one of you guys have a look at this one as well?

davidhorak (approved):
Have done another pass, LGTM

PR #223, opened 2022-03-09
Simplify does not seem to work when the value is an object 

lilyvanoekel:
# Description
Simplification of an expression like `['PRESENT', '$a']` was failing when the value of `$a` in the context is an object. This was due to the check here for `isEvaluable`:

https://github.com/briza-insurance/illogical/blob/master/src/expression/comparison/index.ts#L61

`isEvaluable` considered any Object that's not null or an array to be an `Evaluable`. To fix the issue I made the check more strict in `src/common/type-check.ts`

I've also updated the type of `Result` to indicate a `Result` can be an object

### Test Coverage
- [x] unit

#### References
- Relates to `Bug: Briza Air not rendering Briza API questions`


undees (approved):
🙌🏻 Love it!

lilyvanoekel:
> @lilyvanoekel don't forgot to bump the version (package json) and to update the change log.

5cafb777fcf41bea599d988260202d2ad3ee66b1

PR #231, opened 2022-06-23
Reference performance improvements

luiz290788:
# Description

- Exposes parser default options. This is required to be able to call simplify on an Evaluable directly.
- Exposes isEvaluable
- Implements several performance improvments for Reference.

### Test Coverage
- [x] unit


rpedrosanto (approved):

undees (approved):
🚀

PR #243, opened 2024-07-31
Add Arithmetic support

atoledo:
# Description

- Add Arithmetic Evaluable
  - [x] Sum
  - [x] Subtraction
  - [x] Division
  - [x] Multiplication
- [x] Validate unexpected root operators on parse 
- [x] Simplify arithmetic conditions within other comparison expressions
- [x] Test/support nested arithmetic
- [x] 100% coverage on affected parts. But we still have missing coverage on other areas.

## Note
 - I'll cut the version on a different PR when ready to publish

### Test Coverage
- [x] unit

#### References
- Closes ```GUS-3834```


package-lock.json
```diff
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "@briza/illogical",
-      "version": "1.5.6",
+      "version": "1.5.9",
```
atoledo, line 9: Will cut a new version at the end. This is just the result of npm install and something that was missed previously.

src/common/type-check.ts
```diff
@@ -5,7 +5,7 @@ import { Evaluable, Result } from './evaluable'
  * @param value Tested value.
  */
 export function isNumber(value: Result): value is number {
-  return typeof value === 'number' && isFinite(value)
```
rthreei, line 8: Is this potentially a change in contract?

src/expression/arithmetic/__test__/unit/divide.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    [NaN, operand(0), operand(0)],
+    [2.3333333333333335, operand(2.8), operand(1.2)],
+    [2.0999999099999913, operand(2.333333), operand(1.111111)],
+    [4, operand(0.4), operand(0.1)],
+    [1, operand(1.333), operand(1.333)],
+  ]
+
+  describe('evaluate', () => {
+    test.each(testCases)(
+      'that %p is the result of divideing %p',
```
rthreei, line 34: sp. `divideing`

src/expression/arithmetic/__test__/unit/divide.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    it('should throw if operator symbol is not mapped', () => {
+      expect(() =>
+        new Divide(new Value(10), new Value(20)).serialize({
+          ...defaultOptions,
+          operatorMapping: new Map<symbol, string>(),
+        })
+      ).toThrowError()
+    })
+  })
+})
```
rthreei, line 106: Any expectations for percision?  For cases like:

Screenshot 2024-08-02 at 11 47 33 AM

rthreei (commented):
I think you need update [the docs](https://github.com/briza-insurance/illogical/blob/master/readme.md#working-with-expressions)?

src/common/type-check.ts
atoledo: I thought about it, but I don't think so. 
* Infinity didn't seem to be actively supported. There were no tests for it and not documented.
* Although it's possible to express Infinity in an illogical condition created in a JS array, it will not work while being transported through JSON (and potentially other holders).
image
Seems that it gets converted to `null` instead of throwing error
* Also, before this implementation I don't really see how Infinity would be generated in runtime

Maybe I need to validate if it's possible that the simplify would render Infinity, which would be a problem

src/expression/arithmetic/__test__/unit/divide.test.ts
atoledo: True... sum has the same weird behavior as subtraction. I'll apply the same solution. I think it's worth to adjust this to prevent unexpected conditions not matching due to this rounding problem.

On multiplication and division I don't think it's that needed since those are expected to produce decimais. Wdyt?

src/common/type-check.ts
atoledo: Interesting, `comparison` simplify was never expecting that a single operand could be simplified. I should code this, which would then expose the problem above. I can use the isFinite check to prevent the issue.

src/expression/arithmetic/__test__/unit/divide.test.ts
rthreei: > On multiplication and division I don't think it's that needed since those are expected to produce decimais. Wdyt?

I agree.

rthreei (commented):
Docs need to be updated.  Otherwise looks good.

src/expression/arithmetic/__test__/unit/multiply.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    [6000, operand(10), operand(20), operand(30)],
+    [-200, operand(10), operand(-20)],
+    [-200, operand(-10), operand(20)],
+    [200, operand(-10), operand(-20)],
+    [0, operand(10), operand(0)],
+    [-0, operand(0), operand(-10)],
+    [0, operand(0), operand(0)],
+    [3.36, operand(2.8), operand(1.2)],
+    [2.592591962963, operand(2.333333), operand(1.111111)],
+    [0.04000000000000001, operand(0.4), operand(0.1)],
```
countcain, line 28: is this really expected?

countcain: the max decimal places of a multiply result is the number of left's decimal places + the number of right's
```
> 0.13*0.4
0.052000000000000005
> 0.123*0.234
0.028782000000000002
```
the first result should be 0.052
the 2nd should be 0.028782

atoledo: I can apply the same concept here, but thought it was overkill and unnecessary. See [this](https://github.com/briza-insurance/illogical/pull/243#discussion_r1702175758).

I don't think it's likely that a division and multipliaction will be followed by an Equality. Although that's really specific to our context.

I can work on it

atoledo:
> Docs need to be updated. Otherwise looks good.

Was wondering if I should tag this work as breaking change in the README (I'll cut the actual version in another PR). The update wouldn't break a system that owns the creating of its conditions, so not sure.

rthreei:
Do we have the following test cases?

```
["AND", ["+", 1, -1], ["+", ["-", 1, 1], 1]]]

["AND", ["+", 1, 1], ["-", 2, 1]]

["+", ["AND", ["==", 1, 1]], 1]
```

atoledo:
> Do we have the following test cases?
> 
> ```
> ["AND", ["+", 1, -1], ["+", ["-", 1, 1], 1]]]
> 
> ["AND", ["+", 1, 1], ["-", 2, 1]]
> 
> ["+", ["AND", ["==", 1, 1]], 1]
> ```

Good point, 1 and 2 are not being handled yet and they shouldn't be allowed. 3 is correctly rejected as invalid expression.

rthreei:
> Good point, 1 and 2 are not being handled yet and they shouldn't be allowed.

Curious why you think they shouldn't be allowed?

atoledo:
> > Good point, 1 and 2 are not being handled yet and they shouldn't be allowed.
> 
> Curious why you think they shouldn't be allowed?

Logical expressions should have operands that resolve into a boolean, not other Operands. That's my understanding.

`['AND', 1, 2]` is parsed into `Collection`, not an `And` expression. Weird that currently it doesn't throw `invalid expression` Error, but also doesn't work as we would expect. Seems a "bug".

countcain (approved):

countcain (approved):

PR #245, opened 2024-08-07
npm pkg fix + gitignore

atoledo:
# Description
- npm pkg fix
- gitignore to prevent npm configs to be versioned
- Not worth a release on its own

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   },
   "repository": {
     "type": "git",
-    "url": "git@github.com:briza-insurance/illogical.git"
+    "url": "git+ssh://git@github.com/briza-insurance/illogical.git"
   },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -29,7 +29,7 @@
   },
   "repository": {
     "type": "git",
-    "url": "git@github.com:briza-insurance/illogical.git"
+    "url": "git+ssh://git@github.com/briza-insurance/illogical.git"
```
countcain, line 32: does the new npm version require this change?

countcain (approved):

package.json
atoledo: Don't think so. NPM already made this change before publishing

PR #248, opened 2025-07-03
GUS-5805 PoC for unsafe simplify

atoledo:
# Description
 - PoC for unsafe simplify
 - [ ] Extract operator simplify logic ??
 - [ ] Tests

#### References
GUS-5805

### DIFF
--- a/src/unsafe/simplify.ts
+++ b/src/unsafe/simplify.ts
@@ -1,0 +1,231 @@
+import {
+  isEvaluable,
+  OPERATOR_AND,
+  OPERATOR_EQ,
+  OPERATOR_IN,
+  OPERATOR_NE,
+  OPERATOR_NOT,
+  OPERATOR_OR,
+  OPERATOR_OVERLAP,
+} from '..'
+import { Context } from '../common/evaluable'
+import { isBoolean, isString } from '../common/type-check'
+import { Reference } from '../operand/reference'
+import { Input } from '../parser'
+import { Options } from '../parser/options'
+
+const istrueResult = (value: Input) => value === true
+
+export const unsafeSimplify = (
+  context: Context,
+  opts: Options,
+  strictKeys?: string[],
+  optionalKeys?: string[]
+) => {
+  const simplifyInput = (input: Input): Input | boolean => {
+    // Value or Reference
+    if (!Array.isArray(input)) {
+      // Reference
+      if (isString(input) && opts.referencePredicate(input)) {
+        const result = new Reference(opts.referenceTransform(input)).simplify(
+          context,
+          strictKeys,
+          optionalKeys
+        )
+
+        if (isEvaluable(result)) {
+          return result.serialize(opts)
+        }
+        if (isBoolean(result)) {
+          return result
+        }
+        // It should have been a boolean or an Evaluable, but just in case
+        // fallback to returning the input.
+      }
+      // Value
+      return input
+    }
+
+    const [operator, ...operands] = input
+
+    switch (operator) {
+      // Logical operators
+      case opts.operatorMapping.get(OPERATOR_AND): {
+        const simplifiedOperands: Input = []
+        for (const operand of operands) {
+          const simplification = simplifyInput(operand)
+          if (isBoolean(simplification) && !istrueResult(simplification)) {
+            // Short-circuit for AND
+            return false
+          }
+          simplifiedOperands.push(simplification)
+        }
+
+        const simplified = simplifiedOperands.reduce<boolean | Input[]>(
+          (result, childResult) => {
+            if (result !== false) {
+              if (isEvaluable(childResult)) {
+                if (isBoolean(result)) {
+                  return [childResult]
+                }
+                return [...result, childResult]
+              }
+              if (!childResult) {
+                return false
+              }
+            }
+            return result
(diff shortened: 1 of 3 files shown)


atoledo:
With this current version:
* illogical.simplify no answers: ~42ms
* illogical.simplify with answers: ~30ms
* illogical.unsafeSimplify no answers: ~30ms
* illogical.unsafeSimplify with answers: ~1ms

`no answers` means no short-circuit happens, so all operands were traversed. Worst case.
`with answers` means I gave the answer that applies to the first operand in the first OR. Optimal case.

NOTE: I fixed a "bug" in OVERLAP simplification so both implementations could return the same results.

src/__test__/unit/index.test.ts
```diff
@@ -301,4 +306,189 @@ describe('Condition Engine', () => {
...
+        ['OVERLAP', ['$Ref1', '$Ref2'], [1, 2, 3]],
+        ['OVERLAP', ['$Ref1', '$Ref2'], [1, 2, 3]],
+        {
+          Ref1: 4,
+        },
+        undefined,
+        undefined,
+      ],
+      [
+        true,
```
atoledo, line 456: The bug in Illogical was here. This was simplifying to `['OVERLAP', ['$Ref1', '$Ref2'], [1, 2, 3]]` even if an answer to $Ref1 was provided, which satisfied the expression.

src/operand/collection.ts
```diff
@@ -38,11 +42,13 @@ export class Collection extends Operand {
     for (const item of this.items) {
       const simplifiedItem = item.simplify(...args)
       if (isEvaluable(simplifiedItem)) {
-        return this
+        continue
```
atoledo, line 45: This was the root cause of the bug. The Collection wasn't returning any `Result`s if any reference was missing.

src/unsafe/simplify.ts
```diff
@@ -0,0 +1,272 @@
...
+  }
+
+  if (Array.isArray(value) && value.some(isUndefined)) {
+    return undefined
+  }
+
+  // TODO Confirm below
+  // Having an Object as a Result seems like a previous mistake. Not handling
+  // it yet.
+  return isNotObject(value) ? (value as Input) : undefined
```
atoledo, line 39: Will troubleshoot this next. I think the previous developer got confused with something. We may have an answer as an Object, but I don't think the result can be an Object...

It's kind of expected that the Reference will point to a specific property... not the whole object...

Answer
```
{ Address1: { region: "NY" } }
```

### Valid
Condition
```
["==", "$Address1.region", "NY"] => true
```

### Invalid - the mistake
```
["==", "$Address1.region", "{ "region": "NY" }"] => true
```

Am I missing something?

src/__test__/unit/index.test.ts
atoledo: Yes, many cob conditions are like that.

But we don't have this issue in Core API - quoting flow due to the use of optionalKeys

Maybe this issue can be reproduced in GET /questions. Not sure

src/unsafe/simplify.ts
rthreei: Do we have any conditions that resemble this form?  Probably hard to search for, but if none come to mind, then fixing this should safe.

src/__test__/unit/index.test.ts
atoledo: Reverted and found a way to produce the same results. We can fix this if it becomes a problem in the future.

src/operand/collection.ts
atoledo: I reverted this change since it was causing a bug on the negative case:

`['OVERLAP', ['$Ref1', '$Ref2'], [1, 2, 3]]` with answers as `{ Ref1: 4 }` was also resulting into TRUE

But this feature is what made the Logical short-circuit work, so reverting it caused most of the performace gain be lost.

With [a8fd85d](https://github.com/briza-insurance/illogical/pull/248/commits/a8fd85dc8aee7f9897cca8a15f6a11693fc0f16a) I kept the feature in the unsafe version, while not changing current behavior of the core. Even when not providing strictKEys/optionalKeys.

@rthreei @countcain lmk your thoughts

src/__test__/unit/index.test.ts
```diff
@@ -301,4 +311,942 @@ describe('Condition Engine', () => {
...
+      // AND
+      [
+        false,
+        ['AND', ['==', '$Ref1', 1], ['==', 1, 2], ['==', '$Ref1', 1]],
+        {},
+        undefined,
+        undefined,
+      ],
+      [
+        ['AND', ['==', '$Ref1', 1], ['==', '$Ref1', 1]],
```
countcain, line 375: so simplify doesn't deduplicate?

countcain (approved):

src/__test__/unit/index.test.ts
atoledo: Nope, simplify doesn't "compact" conditions

src/unsafe/simplify.ts
atoledo: Created https://linear.app/briza/issue/GUS-5868/illogical-support-for-objects-as-answers, but honestly this seems way low priority.

PR #276, opened 2026-04-26
add more test for evaluable

luiz290788:
# Description

- adds more tests for evaluable


rthreei (approved):
Nice, did these tests surface any bugs?

luiz290788:
> Nice, did these tests surface any bugs?

Not really but because the simplify tests I created with the same idea found some issues, I also decided to created some more tests for evaluate.

PR #277, opened 2026-04-26
add type assert eslint rule

luiz290788:
# Description

- add type assert eslint rule

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -22,7 +22,9 @@ export default defineConfig([
       parser: tsParser,
       ecmaVersion: 2018,
       sourceType: 'module',
-      parserOptions: {},
+      parserOptions: {
+        projectService: true,
+      },
     },
 
     plugins: {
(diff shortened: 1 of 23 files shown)


eslint.config.js
```diff
@@ -76,6 +78,8 @@ export default defineConfig([
       'no-unused-vars': 'off',
       '@typescript-eslint/no-unused-vars': ['error'],
 
+      '@typescript-eslint/no-unsafe-type-assertion': 'error',
```
rthreei, line 81: 👍

src/operand/value.ts
```diff
@@ -53,6 +53,7 @@ export class Value extends Operand {
    * {@link Evaluable.serialize}
    */
   serialize(): Input {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion
     return this.value as Input
```
rthreei, line 57: This will eventually be removed once we do away with the classes?

rthreei (approved):

src/operand/value.ts
luiz290788: Yes, it will. I tried as much as possible not to ignore the rules but this case I couldn't. And as you mentioned, it will eventually be removed.

atoledo (approved):

PR #280, opened 2026-05-14
add batch evaluator

luiz290788:
## Changes

- add batch evaluator
- [x] Add tests


src/batch/evaluator.ts
```diff
@@ -0,0 +1,286 @@
...
+  }
+
+  /**
+   * Subscribe to change notifications.
+   * Callback fires only for expressions whose result changed.
+   *
+   * @param callback — Called with list of changed expressions
+   * @returns Unsubscribe function
+   */
+  onChange(callback: ChangeCallback): () => void {
```
atoledo: What subscribes these callbacks?

src/batch/evaluator.ts
```diff
@@ -0,0 +1,286 @@
...
+    if (options.options) {
+      for (const key of Object.keys(options.options)) {
+        if (key in this.opts) {
+          Reflect.set(this.opts, key, Reflect.get(options.options, key))
+        }
+      }
+    }
+
+    const expressionsMap = new Map<string, ExpressionInput>()
+    for (const [name, expr] of Object.entries(options.expressions)) {
```
atoledo, line 77: What defines the expression names? The caller?

Shouldn't there be a validation to ensure they are unique?

src/batch/evaluator.ts
```diff
@@ -0,0 +1,286 @@
...
+  /** Cached results from the last evaluation */
+  cachedResults: Record<string, Result>
+  /** Change callbacks registered via onChange() */
+  onChangeCallbacks: ChangeCallback[]
+}
+
+/**
+ * BatchEvaluator — compiles many expressions once, evaluates incrementally.
+ */
+export class BatchEvaluator {
```
atoledo, line 53: So this idea is to pass the whole execution context (Let's say all the question's conditions at once) and evaluate against a single set of answers?

Does this handle the cases where dependencies come out of order? I.e. passes Q1, Q2, Q3, Q4 conditions. Where Q2 depends on Q1, but Q3 depends on Q4.

rthreei:
Would be interesting to see performance metrics.

Also, would be useful to benchmark batch evaluation versus individual evaluation.  For example:
- given say 1,000 fields with varying conditions
- when 1 field changes
- how long does it take to "batch evaluate"
- vs. evaluating all conditions individually

types/index.d.ts
```diff
@@ -55,5 +58,29 @@ declare class Engine {
...
+     * @param {BatchEvaluatorOptions} options — Expressions map and optional parser options
+     * @returns {BatchEvaluator}
+     *
+     * @example
+     * ```typescript
+     * const batch = engine.createBatchEvaluator({
+     *   expressions: {
+     *     isActive: ['==', '$status', 'active'],
+     *     isPremium: ['==', '$tier', 'premium'],
+     *   },
```
rthreei, line 76: Can expressions be modified after instantiation?  Probably not, and I can't think of a valid use case needing that.

types/index.d.ts
```diff
@@ -55,5 +58,29 @@ declare class Engine {
...
+    /**
+     * Create a BatchEvaluator for evaluating multiple expressions with shared resources.
+     *
+     * The batch evaluator compiles all expressions once, shares refs/consts across them,
+     * and supports incremental evaluation based on trusted dirty keys.
+     *
+     * @param {BatchEvaluatorOptions} options — Expressions map and optional parser options
+     * @returns {BatchEvaluator}
+     *
+     * @example
```
rthreei, line 70: Would be good to add to docs to /specs folder.

src/batch/evaluator.ts
luiz290788: I don't know if I understand what you mean here. In the point of view of illogical, there is no dependency between conditions.

src/batch/evaluator.ts
luiz290788: Nothing, it is just a public API that probably doesn't need to exist. I will remove it.

src/batch/evaluator.ts
luiz290788: Yes, the expression ids are defined by the caller. I don't know if I understand the concern about uniqueness here. The expressions are provided as part of the `expressions` property in the `BatchEvaluatorOptions` and that is a `Record<string, ExpressionInput>` I don't see how that record would have multiple entries with the same key.

types/index.d.ts
luiz290788: They cannot. I also don't see a use that where that would be useful. If we find a new use case where that is useful, I would rather implement a remove and add of expressions.

specs/batch-evaluation.md
```diff
@@ -0,0 +1,395 @@
...
+batch.onChange((changes) => {
+  for (const change of changes) {
+    if (change.name === 'canSubmit') {
+      submitButton.disabled = !change.current
+    }
+    if (change.name === 'needsVerification') {
+      showVerificationBadge(change.current)
+    }
+  }
+})
```
rthreei: Should this (and related onChange documentation) be removed?  Since the implementation was removed.

benchmark/report-batch.md
```diff
@@ -0,0 +1,102 @@
...
+| medium-complex | 13 | 99.7µs | 39.7µs | individual 0.40x |
+| large-complex | 22 | 639.2µs | 52.7µs | individual 0.08x |
+
+---
+
+## Key Findings
+
+1. **Batch evaluation wins when many expressions are affected** — the shared compilation and dependency graph help amortize overhead.
+2. **Individual evaluation wins when few expressions change** — JIT-optimized bytecode is very efficient for small sets, and batch has overhead from dependency graph lookup + interpreter dispatch.
+3. **For the "1,000 fields, 1 field changes" scenario:** individual evaluation is significantly faster than batch for incremental updates.
```
rthreei, line 90: This is both interesting and unexpected.

For clarification, evaluating 1,000 expressions individually when only 1 field changes is faster than batch evaluation?

Why is that?  Presumably batch evaluation would only evaluate expressions that reference the 1 changed field.  What am I missing?

src/batch/evaluator.ts
atoledo: I mean the use of strictKeys (and maybe optionalKeys) that we have in our simplify process, which does imply some sort of dependency.

If each condition belongs to a reference and these references can exist in other conditions, there is a proper order that the conditions need to be processed. That's what I'm trying to think if it would apply here. Otherwise we would have group into tiers, then run each tier in a batch.

luiz290788: The evaluation process doesn't know anything about strict keys and optional keys. And there is no explicit implication in the illogical side, we should not assume anything.

atoledo: Never realized that those 2 were just for simplification and not evaluation. Makes sense then. My thought here isn't applicable. Thanks

specs/batch-evaluation.md
luiz290788: Done!

benchmark/report-batch.md
luiz290788: I found two problems, we were compiling the expressions on every evaluate and we could compile once and then reuse.
Also, we were comparing it wrong, we were including the compile time in the comparison, but the compile time of all 1000 expressions versus the compile time of 15 expressions. The batch execution of a single property change that caused 15 expressions to be re-evaluated, was compiling 1000 vs the baseline that was compiling 15 expressions.

rthreei: That sounds positive!  What are the "key findings" now?

src/batch/__tests__/unit/dependency-graph.test.ts
```diff
@@ -0,0 +1,140 @@
...
+          '{region}',
+        ]
+        const graph = buildDependencyGraph(
+          expressions,
+          sharedRefs,
+          refKeys,
+          defaultOptions
+        )
+
+        console.log(JSON.stringify(Array.from(graph.entries()), null, 2))
```
countcain: remove?

src/batch/__tests__/unit/dependency-graph.test.ts
```diff
@@ -0,0 +1,140 @@
...
+
+        // simple string ref: status -> expr1 (deduplicated to 1 entry)
+        assert.deepEqual(graph.get('status'), [
+          { exprName: 'expr1', refIdx: 0 },
+        ])
+
+        // multi-key ref produces entries for all segment keys
+        assert.deepEqual(graph.get('user'), [{ exprName: 'expr2', refIdx: 1 }])
+        // TODO: Verify in v2 what happens if a key is both a top-level key and part of a multi-key ref
```
atoledo, line 91: - [x] Need to verify this, something feels wrong to mark the property as a reference

PR #282, opened 2026-05-26
Improve parse test

atoledo:
# Description

- The current tests were basically testing the same case of  `[null]`.
- Fixing it to parse the actual operators rather than the Symbol of the operators.
- [ ] Arithmetic is not yet catching runtime errors

### Test Coverage
- [x] unit

#### References
- None


countcain:
I will add Arithmetic ones in my PR.

countcain (approved):

PR #287, opened 2026-06-02
GUS-6756 Update GHA and author

atoledo:
# Description

- Update GHA to run on node 24 and author

### Test Coverage
- [ ] unit

#### References
GUS-6756


.github/dependabot.yml
atoledo, line 1: all the ignored versions were for older versions, and Dependabot won't need to skip those minor/patch bumps on outdated majors anymore.

rthreei (approved):

PR #289, opened 2026-06-02
fix simplify

countcain:
# Description

add test cases for simplify

### Test Coverage

- [ ] unit

#### References

gus-6987


conditions/simplify-conditions/224-optinal-keys-2.json
```diff
@@ -0,0 +1,56 @@
...
+            "561410-14"
+          ]
+        ]
+      ]
+    ]
+  ],
+  "context": {
+    "A": "561410-14"
+  },
+  "expected": ["==", "$B", "yes"],
```
countcain: Both implementations return false here, which I think is wrong.

conditions/simplify-conditions/223-optional-keys.json
```diff
@@ -0,0 +1,56 @@
...
+            "561410-14"
+          ]
+        ]
+      ]
+    ]
+  ],
+  "context": {
+    "A": "561410-14"
+  },
+  "expected": ["==", "$B", "yes"],
```
countcain, line 53: oop returns as expected but bytecode return true

conditions/simplify-conditions/225-optional-keys-3.json
```diff
@@ -0,0 +1,54 @@
...
+            "561410-14"
+          ]
+        ]
+      ]
+    ]
+  ],
+  "context": {
+    "A": "561410-14"
+  },
+  "expected": ["==", "$B", "yes"]
```
countcain, line 53: oop returns as expected, bytecode returns true, which is wrong.

conditions/simplify-conditions/233-complex-composition.json
```diff
@@ -0,0 +1,24 @@
...
+  "context": {
+    "A": 14,
+    "E": 5
+  },
+  "expected": [
+    "OR",
+    ["==", "$B", "yes"],
+    ["==", "$D", 1],
+    ["OR", ["==", "$C", "yes"], ["==", "$B", "no"]]
+  ]
```
countcain, line 35: oop returns expected, bytecode returns ["==", "$B", "no"], which is wrong.

PR #291, opened 2026-06-06
fix(simplify): support arbitrarily deep AND/OR nesting via OP_ENTER_SCOPE

luiz290788:
## Summary

- Fixes `simplify()` returning wrong results for conditions with 3+ levels of nested AND/OR/NOR. The bug caused outer residuals to be silently discarded when an inner AND/OR scope executed.
- Adds 19 new simplify test cases (11 from PR #289, 8 new) that cover the failure modes identified in GUS-6987 QA.
- Zero performance regressions; pre-existing minor regressions on `expression-simple-ne` and `expression-medium-or` are also resolved.

## Root cause

The simplifier used a single flat spill buffer shared across all nesting depths. A `lastJumpOp` heuristic cleared it whenever the jump type changed (AND→OR or OR→AND), which wiped outer-scope residuals the moment an inner AND/OR scope started. This produced correct results only for one level of nesting.

## Fix: `OP_ENTER_SCOPE` (opcode 58)

A new no-operand opcode is emitted by the compiler at the start of every `emitShortCircuit` block. In the simplifier interpreter it pushes the current `spillTop` onto a scope stack. Each `OP_AND`/`OP_OR`/`OP_NOR` pops its own base, drains only `spillBuf[base+1..spillTop]` (entries belonging to that scope), then restores `spillTop = base` — leaving outer scopes' residuals intact. The old `lastJumpOp` clearing logic is removed entirely.

The evaluate interpreter and debugger-vm treat `OP_ENTER_SCOPE` as a no-op.

## Performance (bytecode vs OOP after fix)

| Benchmark | Faster | Slower |
|-----------|--------|--------|
| Synthetic simplify (87 cases) | **87** | 0 |
| Sample simplify (595 cases) | **595** | 0 |
| Synthetic evaluate (106 cases) | **106** | 0 |
| Sample evaluate (765 cases) | **765** | 0 |

Worst-case (least improved) simplify case is still **+25% faster** than OOP. Before-vs-after the fix shows no regressions on sample conditions and slight improvements on synthetic (removing the `lastJumpOp` branch from the JUMP hot path offsets the one new `OP_ENTER_SCOPE` push per AND/OR block).

## Test coverage

- 223–232: copied from PR #289 (`strictKeys`/`optionalKeys` cases)
- 233: PR #289 complex composition case (`{A:14, E:5}`)
- 234: `OR(unknown, AND(known-true, unknown))` — outer residual lost entering inner AND
- 235: `AND(unknown, OR(known-true, unknown))` — outer residual lost entering inner OR
- 236: `OR(unknown, AND(known-true, unknown), unknown)` — 3-operand OR with middle AND
- 237: `OR(unknown, OR(unknown, unknown))` — inner OR stealing outer OR's residuals
- 238: `AND(unknown, AND(unknown, unknown))` — inner AND stealing outer AND's residuals
- 239: 3-level `OR → AND → OR` nesting
- 240: 4-level `OR → AND → OR → AND` nesting
- 241: Exact case from the GUS-6987 QA PDF report (`{A:15, E:5}`)

Closes #289
References GUS-6987


src/bytecode/opcodes.ts
```diff
@@ -65,3 +65,9 @@ export const OP_NOR = 55 // next: N — marker for NOR with N operands
 export const OP_IN_SCAN_REFS_CONST = 56
 // next: N, ref0..refN-1, constIdx — resolve each ref inline, check ∉ consts[constIdx], no stack alloc
 export const OP_NOT_IN_SCAN_REFS_CONST = 57
+// Marks the entry of a short-circuit AND/OR/NOR scope.
```
rthreei, line 68: Does XOR and NOT need to be considered?  Or are those working as expected, and not effected by the bug?

luiz290788: they are not affected.

NOT -> it doesn't create a new scope because it is a unary operator.
XOR -> it is compiled as chain and doesn't have the short circuit implemented.

src/bytecode/opcodes.ts
atoledo, line 10: Can `OP_PUSH_REF` be removed? Marked as legacy and there are no usages of it

countcain (dismissed):

countcain (approved):

PR #294, opened 2026-06-11
add openssf scorecard

countcain:
closes gus-7075

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -1,7 +1,3 @@
-# This workflow uses actions that are not certified by GitHub. They are provided
-# by a third-party and are governed by separate terms of service, privacy
-# policy, and support documentation.
-
 name: Scorecard supply-chain security
 on:
   # For Branch-Protection check. Only the default branch is supported. See
@@ -12,7 +8,7 @@ on:
   schedule:
     - cron: '19 6 * * 3'
   push:
-    branches: [ "master" ]
+    branches: ['master']
 
 # Declare default permissions as read only.
 permissions: read-all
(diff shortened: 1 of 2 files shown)


.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,74 @@
...
+      security-events: write
+      # Needed to publish results and get a badge (see publish_results below).
+      id-token: write
+      # Uncomment the permissions below if installing in a private repository.
+      # contents: read
+      # actions: read
+
+    steps:
+      - name: 'Checkout code'
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
atoledo: Why not receive updates automatically?
```suggestion
        uses: actions/checkout@v6
```

atoledo: Actually, this is in v6 already

atoledo: same for the other actions below

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,54 @@
...
+      security-events: write
+      # Needed to publish results and get a badge (see publish_results below).
+      id-token: write
+    steps:
+      - name: 'Checkout code'
+        uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+      - name: 'Run analysis'
+        uses: ossf/scorecard-action@v2
```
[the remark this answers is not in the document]
atoledo: Is this issue specific to `ossf/scorecard-action` ? We won't receive updates if we bind to a specific commit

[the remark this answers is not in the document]
rthreei: I think so because it is third-party and not managed by Github.

atoledo (approved):

PR #303, opened 2026-06-22
GUS-7152 Add fast-check for fuzzing tests

atoledo:
# Description

- Add fast-check for fuzzing tests
- [x] All operators are being tested
- Discuss about the idea of the fuzz test in test.yml failing CI or not
- Discuss about nightly / weekly CI
- [x] The generator isn't perfect and not sure it should produce "bad" data
- [x] Improve context for Evaluate / Simplify tests
- [x] Added some reporting improvement, but up for discussion
  - Shows how many runs were successful VS errors
  - Any unexpected error would fail CI
    - Doesn't mean that the errors are fully correct, but aiming to reduce noise.
  - Then calculates how many conditions had each operator, to ensure we don't break coverage
  - Added some samples for each operator, so we can quickly review if the generator still produces what we expect

### Test Coverage
- [x] unit

#### References
GUS-7152


.github/workflows/fuzz-weekly.yml
```diff
@@ -0,0 +1,32 @@
...
+      - name: Run Build
+        run: npm run build
+      - name: Run Deep Fuzzing (100k runs)
+        run: npm run test:fuzz:deep
+      - name: Upload Fuzzing Report
+        uses: actions/upload-artifact@v7
+        if: always() # Ensure it uploads even if tests crash
+        with:
+          name: fuzz-reports
+          path: ./fuzz-*-report.json
```
atoledo: Not sure how valuable is this, or if we want to save a full list of tested expressions.

.github/workflows/test.yml
```diff
@@ -25,5 +25,7 @@ jobs:
         run: npm run lint
       - name: Run test suite
         run: npm t
+      - name: Run quick fuzzing
+        run: npm run test:fuzz || true
```
rthreei: Why always have it pass?  Is it flakely?

atoledo: Until I decided to have `expectedErrors` it was. But now it isn't anymore

It is producing ~5% errors, which I guess is a good thing.

I'm planning on searching for other expected errors that are not being observed in results and see what I can do to make them active.

src/__test__/fuzz/base.fuzz.test.ts
```diff
@@ -0,0 +1,272 @@
...
+      result?: unknown
+      error?: string
+    }
+  >
+}
+
+describe('Fuzzing Tests', () => {
+  const engineOOP = new Engine({ evaluator: 'oop' })
+
+  const EXPECTED_ERRORS = new Set<string>([
```
atoledo, line 32: > The generator isn't perfect and not sure it should produce "bad" data

So the current approach is trying to get generators not to throw, so it minimizes structural errors. These `EXPECTED_ERRORS` are the ones thrown in runtime since the generator sometimes use a Ref value that is incompatible.

countcain (approved):

PR #306, opened 2026-06-24
Add note on non-present operands for arithmetic expressions

matthewhelmke:
# Description

Added a note that covers when an arithmetic expression has a missing operand. Before we merge, do we want to add anything that covers the "weird" edge case where evaluating `["==",["+","$Ref1",0],["+",0,"$Ref2",0]]` with an empty context results in TRUE (because both arithmetic expressions result in FALSE due to both refs being absent)?

I added content to cover the main case. For this edge case, we would also want to note that it is possible to overcome the expected behavior by appending PRESENT expressions on each Ref merging with an AND.

But, 1) how likely is this and 2) how detailed do we want to go with edge cases? I'm here and willing, but is it worth what my time costs? I'm not a good judge of that (because I'm really happy for you to pay for my time).

@atoledo please ping me if you think it's worth a quick video chat and I'm happy to schedule one!!

### DIFF
--- a/specs/arithmetic-expressions.md
+++ b/specs/arithmetic-expressions.md
@@ -9,7 +9,7 @@ An arithmetic expression combines numeric values, variables, and arithmetic oper
 
 Arithmetic Expressions are not supported as root level expressions since those must evaluate to a boolean. But arithmetic expressions can be used nested within [Comparisson Expressions](./comparison-expressions.md).
 
-> Note: Any non-present operand is discarded. This means that if a reference is missing, the expression will return FALSE. Also, all operands that ARE present must be numbers, otherwise an error is thrown.
+> Note: Any non-present operand is discarded. This means that if a reference is missing, the expression will return FALSE. Also, all operands that ARE present must be either numbers or an ISO date string followed by valid date durations, otherwise an error is thrown.
 
 ## Division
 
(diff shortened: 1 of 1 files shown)


atoledo:
That's a good point. Probably better to focus on being objective and not document each possible relation from one expression with another, which would make the documentation too extensive.

People with knowledge on each related expressions would be able to solve that edge case.

specs/arithmetic-expressions.md
```diff
@@ -9,6 +9,8 @@ An arithmetic expression combines numeric values, variables, and arithmetic oper
 
 Arithmetic Expressions are not supported as root level expressions since those must evaluate to a boolean. But arithmetic expressions can be used nested within [Comparisson Expressions](./comparison-expressions.md).
 
+> Note: Any non-present operand is discarded. This means that if a reference is missing, the expression will return FALSE. Also, all operands that ARE present must be numbers, otherwise an error is thrown.
```
atoledo: I forgot about one detail.

There was a recent change to allow Arithmetic operations on Dates as well. There are some examples below.

```suggestion
> Note: Any non-present operand is discarded. This means that if a reference is missing, the expression will return FALSE. Also, all operands that ARE present must be either numbers or an ISO date string followed by valid date durations, otherwise an error is thrown.
```

matthewhelmke:
@atoledo I've adjusted the note with your updated info. Thanks! Anything else we want to do with this? If not, I'm good for you to merge.

atoledo (approved):

PR #307, opened 2026-06-24
GUS-6987 Side effect found on nested logical

atoledo:
# Description

- Add test cases to reproduce issue `249-and-or-eqs.json`
- Added some other test cases just to document expected outcome
- [x] Review 2.2.3 diff and identify root cause
  - On 2.2.2, the simplification was performing an unwanted compacting on OR EQ -> IN
  - On 2.2.3, the second operand is getting discarded
- [x] Some other edge cases were adjusted and had the optimized logic skip structures that wouldn't be able to be regenerated during simplify
- [x] Fixed debugger to also show this optimized case correctly
- Revert changes, but merging the overall improvements to tests, debugger and benchmark

## Scenario 249-and-or-eqs.json
Expected simplified result
```
[
  "AND",
  ["AND", ["==", "$LossHasAny", "yes"], ["==", "$Loss1Type", "property"]],
  [
    "OR",
    [
      "AND",
      ["==", "$Loss1Type", "property"],
      ["==", "$Loss1ClaimStatus", "closed"]
    ],
    [
      "AND",
      ["==", "$Loss1Type", "property"],
      ["==", "$Loss1ClaimStatus", "declined"]
    ]
  ]
]
```

Bytecode v2.2.2 was compacting the OR EQ into IN
```
[
  "AND",
  ["AND", ["==", "$LossHasAny", "yes"], ["==", "$Loss1Type", "property"]],
  [
    "OR",
    [
      "AND",
      ["==", "$Loss1Type", "property"],
      ["IN", "$Loss1ClaimStatus", ["closed", "declined"]]
    ]
  ]
]
```

Bytecode v2.2.3 is removing the second operand
```
[
  "AND",
  [
    "==",
    "$LossHasAny",
    "yes"
  ],
  [
    "==",
    "$Loss1Type",
    "property"
  ]
]
```

### Test Coverage
- [x] unit

#### References
- Closes ```[ISSUE-ID]```


conditions/evaluate-conditions/157-sum-date-true.json
atoledo, line 1: These 5 test cases are unrelated to the actual issue. But worth documenting.

countcain (dismissed):

rthreei:
Did you run benchmark on the changes?

atoledo:
> Did you run benchmark on the changes?

Results are not good, but just for this type of condition. The rest remaing consistent as it was like expected. I'll check if there is anything to improve or why it performs worse than oop.

Update: the regression wasn't caused by this PR though. It master branch the problematic scenario already has -40% performance.

conditions/to-support/simplify-conditions/249-and-or-eqs.json
atoledo, line 1: These 3 tests are the ones causing issues. The unit tests also covering these were skipped since I reverted the fixes.

countcain (approved):

PR #313, opened 2026-09-03
Fix bytecode OOP mismatch on multi-operand arithmetic with missing refs

luiz290788:
Multi-operand SUM/SUBTRACT (3+ operands) with missing refs produced a different result in bytecode than the OOP evaluator. The multi-operand path wrote to the wrong stack slot; now it writes to the operand slot and resets the stack top, matching OOP. Adds regression tests + sample conditions.

### DIFF
--- a/src/bytecode/interpreter.ts
+++ b/src/bytecode/interpreter.ts
@@ -512,7 +512,8 @@ export function interpret(compiled: CompiledExpression, ctx: Context): Result {
       }
 
       case OP_OR_AND_IN_CONST_2: {
-        // bytecode layout: ref1Idx, ref2Idx, M, (v0, setBIdx0, ref1Op0, ref2Op0), (v1, setBIdx1, ref1Op1, ref2Op1), ...
+        // bytecode layout: ref1Idx, ref2Idx, M,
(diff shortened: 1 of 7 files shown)


rthreei:
@atoledo this appears to address GUS-7155

atoledo (approved):

src/bytecode/interpreter.ts
```diff
@@ -669,9 +670,15 @@ export function interpret(compiled: CompiledExpression, ctx: Context): Result {
           break
         }
         const values: Array<string | number> = new Array(n)
+        // Fixed destination slot for the single result value: the bottom-most
+        // operand slot. The operands occupy [stackTop - n + 1 .. stackTop].
+        // Writing here (instead of to stack[++stackTop]) keeps the result in a
+        // stable slot even when the loop below breaks early on a missing
+        // operand, so the next opcode never reads a stale operand value.
```
rthreei, line 677: While the comment makes sense, it doesn't seem specific to the immediately proceeding lines.

atoledo: Seems ok to be anchored on the resultSlot though. Then pulling the commit diff would help clarify intent.

PR #314, opened 2026-09-03
Fix bytecode simplify dropping operands in nested OR/AND composition

luiz290788:
Bytecode simplify collapsed nested OR/AND structure via the OR_AND_IN merge, which could only reconstruct the flat 2-level case and dropped operands in deeper compositions. For simplify we now skip that merge so the original nested structure is preserved verbatim (evaluate keeps the optimization for performance). Adds regression tests, including arbitrary-depth nesting.

### DIFF
--- a/src/bytecode/compiler.ts
+++ b/src/bytecode/compiler.ts
@@ -163,6 +163,14 @@ export interface CompilerState {
   // Side tables for the simplify interpreter
   overlapRefsEntries: Array<{ pos: number; refIdxs: number[] }>
   directionEntries: Array<{ pos: number; dir: 0 | 1 }>
+  // When true, the compiler preserves the original nested structure for the
+  // simplify interpreter instead of applying the OR(AND(ref1, ref2), …)
+  // OR_AND_IN merge optimization. That optimization collapses branches that
+  // share ref1's value into a single merged ref2 set, which is lossy for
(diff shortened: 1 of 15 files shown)


atoledo (approved):

src/bytecode/compiler.ts
```diff
@@ -163,6 +163,14 @@ export interface CompilerState {
...
   overlapRefsEntries: Array<{ pos: number; refIdxs: number[] }>
   directionEntries: Array<{ pos: number; dir: 0 | 1 }>
+  // When true, the compiler preserves the original nested structure for the
+  // simplify interpreter instead of applying the OR(AND(ref1, ref2), …)
+  // OR_AND_IN merge optimization. That optimization collapses branches that
+  // share ref1's value into a single merged ref2 set, which is lossy for
+  // structural reconstruction — simplify must be able to reproduce the input
+  // verbatim, so it skips the merge and lets the short-circuit path preserve
+  // each branch. evaluate() keeps the optimization for its performance gain.
+  simplify: boolean
```
atoledo, line 173: This ended up causing a problem when Evaluables are parsed individually in order to be cached and later on simplified.

The standard Engine parse is always assuming false. If simplify is called from the Evaluable, it executes over the non-simplified compiled version, producing a different outcome.

PR #319, opened 2026-09-10
GUS-7437 Adjust Evaluable simplify

atoledo:
# Discussion

- The recent simplify fix still reproduced the original issue, and unveiled a bigger discussion. I think we need to set a strong assumption that the Bytecode must be able to reconstruct the original expression.
  - Optimizations must not break it, which means they can only get creative when processing the bytecode and managing the `stack` in a smarter way. Without changing structure.
  - The structure must always allow reconstruction.
  - Which means we need the serialization feature implemented so we are able to add some integrity tests.
  - `toString` also needs to be implemented so we can remove the OOP parse fallback, which is likely dropping performance.

# Description
- Revert simplify fix
- [x] Propose removal of `OP_OR_AND_IN_CONST_2` operator which creates a bytecode that doesn't allow reconstruction. Remove other references if agreed.
- [x] Regenerate types
- [x] Add report

## Future PRs
- Implement `serialize`
- Implement `toString` and remove OOP fallback
- Create integrity test that asserts that an expression can be serialized back into its original form

### Test Coverage
- [x] unit

#### References
- Closes GUS-7437


rthreei:
@atoledo is this ready for review?  Asking because it is in Draft.

atoledo:
The current code is expressing a path I'm proposing, but it isn't ready to be merged. I'd like some feedback from the team on the discussion before continuing.

rthreei:
> The current code is expressing a path I'm proposing, but it isn't ready to be merged. I'd like some feedback from the team on the discussion before continuing.

Looks reasonable to me.

@countcain @luiz290788 ?

rthreei (dismissed):

conditions/simplify-conditions/253-and-ors.json
```diff
@@ -0,0 +1,43 @@
...
+      [
+        "AND",
+        ["==", "$Loss1Type", "property"],
+        ["==", "$Loss1ClaimStatus", "closed"]
+      ],
+      [
+        "AND",
+        ["==", "$Loss1Type", "property"],
+        ["==", "$Loss1ClaimStatus", "declined"]
+      ]
```
guidimas, line 40: [non-blocking] Replace references
```suggestion
      [
        "AND",
        ["==", "$a", "property"],
        ["==", "$b", "declined"]
      ]
```

src/index.ts
guidimas, line 1: [minor] Indirect test case found on [src/__test__/unit/index.test.ts](https://github.com/briza-insurance/illogical/blob/b2c76927bddf07feffda587fcb9117c60e76be13/src/__test__/unit/index.test.ts#L1636-L1656) for removed `OP_OR_AND_IN_CONST_2`:
- `should simplify OR-AND-IN pattern with null literal (Bug 2)`

guidimas (dismissed):
Minor comments, can be addressed on next PR.

src/index.ts
atoledo: what are you proposing here?

rthreei: I think Dimas is suggesting that we remove the test case?  Since it is related to OP_OR_AND_IN_CONST_2.

atoledo: Maybe i can change the commet around it, or just move the test case to the main tests. But don't think we should remove it.

rthreei (approved):

PR #321, opened 2026-09-11
GUS-7441 Add Bytecode serializer and toString

atoledo:
# Description

- Implementing `serializer` for Bytecode
- Implemented `toString` for Bytecode
- Add test for hard assumption that each Opcode implementation must be able to reconstruct the input expression
- Remove OOP fallback from Bytecode parse
- [x] Update types
- [x] Merge and adjust changes in base branch
- [x] Rerun bench reports

### Test Coverage
- [x] unit
- [x] Cover 100% of Opcodes

#### References
GUS-7441


src/bytecode/disassemble.ts
atoledo, line 1: Not really related, but I used this during debugging. I extracted the script part into a tools so it doesn't get built with the app.

atoledo:
I found an optimization in 3.x branch, which might explain the parse performance. 

`getCompiled` is calling `this.parser.parse(exp)`, which is the OOP Parser. Need to port that fix to completely remove dependency on OOP.

src/index.ts
atoledo, line 1: Removed the 2 fallbacks to OOP from Bytecode. This should change the performance results within our API.

The validation was then moved to within the Bytecode compiler. I took the version from `illogical-3` branch and modified/enhanced it based on the new tests that were introduced in master.

benchmark/report-sample-parse.md
```diff
@@ -8,9 +8,9 @@
 
 | | Count |
 |---|---|
-| Faster (>+5%) | 19 |
-| Slower (>-5%) | 34 |
-| Unchanged | 563 |
+| Faster (>+5%) | 616 |
```
rthreei, line 11: Cool!

rthreei (approved):

PR #326, opened 2026-09-21
GUS-7479 Batch evaluator on OOP

atoledo:
# Description

- Add Batch evaluator to support holding the state of multiple expressions
- Remove Bytecode references leftovers
- [x] Add README
- [x] Performance analysis

### Test Coverage
- [x] unit

#### References
GUS-7479


src/batch/batch.ts
```diff
@@ -0,0 +1,248 @@
...
+   *   Context is merged into stored context, all expressions run.
+   *
+   * Mode 2 — With changedKeys: incremental evaluation.
+   *   Context is merged, only expressions affected by changedKeys run.
+   *   The caller guarantees that only these keys actually changed.
+   *
+   * Mode 2 — With empty changedKeys []: no-op, returns cached results.
+   *
+   * @param ctx — Full evaluation context
+   * @param changedKeys — Optional list of keys that changed (trusted by caller)
```
atoledo, line 84: Maybe we should check if `changedKeys` exist in the dependency list. To help callers provide the correct expected key, which in cases of dynamic/property/casting/etc is not trivial to know what is "the key"

rthreei: > Maybe we should check if changedKeys exist in the dependency list.

Throw an error if it doesn't?  Is that the thought?

src/batch/batch.ts
```diff
@@ -0,0 +1,248 @@
...
+   * Merge caller's context into stored context. Handles deletion via
+   * undefined sentinel.
+   */
+  private mergeContext(ctx: Context): void {
+    const map = new Map<string, ContextValue>(
+      Object.entries(this.state.lastContext)
+    )
+
+    for (const [key, newVal] of Object.entries(ctx)) {
+      if (key === '__proto__') {
```
rthreei, line 264: Just curious, what is the purpose of this?

src/batch/batch.ts
atoledo: Yes, listing the available keys that exist in context

src/batch/batch.ts
atoledo: Prototype pollution security issue that CodeQL flagged. And I discovered another bug in illogical due to it.

Currently, we can't have a Reference as `constructor` for example.

I was planning on opening an issue for discussion and see if we want to touch it or leave as-is.

src/operand/reference.ts
```diff
@@ -220,6 +220,10 @@ export class Reference extends Operand {
     return `{${this.key}}`
   }
 
+  getKey(): string {
+    return this.key
```
atoledo, line 224: I'm investigating a bug related to this. This key is not exactly what I wanted it to be.

This key doesn't hold the "ID" of the element. It's the output from the `Options.referenceTransform`, without the type casting. But including array index, dynamic reference, nested object property...

So depending on the condition, we might have different keys that actually point to the same reference...

```
["PRESENT", "$Ref1"]
```
key: Ref1

```
["==", "$Ref1.(Number)", 1000]
```
key: Ref1

```
["==", "$Ref1.prop1", 1000]
```
key: Ref1.prop1

```
["==", "$Ref{index}", 1000]
```
key: Ref{index}
Note: Dynamic keys will also need special handling. Either always getting re-evaluated or with a better comtrol to check if it resolves to something present in the changed keys.

___
This "Root Key" or "Principal Key" concept should be better extracted in the `Reference`

src/batch/batch.ts
```diff
@@ -0,0 +1,248 @@
...
+   *   Context is merged, only expressions affected by changedKeys run.
+   *   The caller guarantees that only these keys actually changed.
+   *
+   * Mode 2 — With empty changedKeys []: no-op, returns cached results.
+   *
+   * @param ctx — Full evaluation context
+   * @param changedKeys — Optional list of keys that changed (trusted by caller)
+   * @returns Record mapping expression names to their Result values
+   */
+  evaluate(ctx: Context, changedKeys?: string[]): Record<string, Result> {
```
atoledo: If the new context is merged with stored context, shouldn't we compute `changedKeys` based on context keys instead? 🤔 

Passing new context values and failing to set the key in changedKeys will produce unintended results.

atoledo: Yeah, some tests are weird because of it

```
      {
        name: 'returns cached results immediately when changedKeys is empty',
        options: {
          expressions: {
            isAdult: ['>=', '$age', 18],
          },
        },
        initial: {
          context: { age: 20 },
        },
        context: { age: 15 },
        changedKeys: [],
        expectedResults: { isAdult: true },
      },
```

The engine can detect age changed and just run correctly

src/batch/batch.ts
atoledo: Decided to remove changedKeys. This exposes to the caller complexities that are not well documented/handled by illogical's Reference.

Better to compute what changed based on provided context.

src/batch/batch.ts
atoledo, line 1: Performance thread

Image

* The initial parsing time is still higher (will break down why, but it's probably the dependency graph calculation)
* Subsequent evaluations are much faster. From ~33ms to ~10ms, due to only evaluating a subset of expressions.

Note: From the 1125 expressions, 711 of them have a condition to the context value that is changing.

rthreei: > to ~10ms

Can this be faster?

rthreei: > The initial parsing time is still higher

This too - can this be faster?

atoledo: On the parsing:

* 15% of the time is the actual parse = ~45ms
* The rest is building the dependency graph = ~266ms

If the Evaluable had a `getReferences()` method, which got populated as the expressions is already being traversed, this would greatly improve.

Image

src/batch/dependency-graph.ts
```diff
@@ -0,0 +1,161 @@
...
+    // Reuse Reference key logic to handle complex keys, casting, and array indices correctly.
+    const reference = new Reference(opts.referenceTransform(expression))
+
+    const key = reference.getKey()
+
+    // Normalize the key extracting the root part before any complex components.
+    // Ideally, Reference operand should provide a method to directly get the
+    // root key, but this is not available yet.
+    const rootKey =
+      key[0] === '[' || key[0] === '.' ? key : key.split(/[.[]/)[0]
```
atoledo: Another point for discussion is this.

It feels wrong to have the Ref key contain the complex elements (array, nested object property, dynamic reference). But at the same time, it is hard to solve for values that are objects, since we could have all complex elements.

How it is currently done in Batch mode is detecting the primary attribute that would be present in the Context. So:
* Ref1 => Ref1
* Ref1.prop2 => Ref1
* Ref1[0] => Ref1

Then basically, any time the value for `Ref1` changes in the context, conditions that use the references above would be recomputed.

Object optimization so that conditions depending on `Ref1.prop2` is only recomputed if `prop2` actually changes was not made.

Now, for Dynamic references, it would be way harder. Would need to create another Map linking placeholders with the key, etc. Currently all Dynamic references are always recomputed.
* Ref{index} => Ref{index}

___

The alternative would be to have `reference.getKey()` return something like:
```
{
  id: "Ref1",
  property: "prop2"
  index: 0
}
```
Basically making any modifier first class.

rthreei: So is there potential for a bug? Or just that performance for Dynamic References will not be as great?

atoledo: just that performance for Dynamic References will not be as great :+1:

src/batch/batch.ts
```diff
@@ -0,0 +1,276 @@
...
+   * Get dependency information.
+   * @returns Map of context key → list of expression names that depend on it
+   */
+  getDependencies(): Map<string, string[]> {
+    const deps = new Map<string, string[]>()
+    for (const [key, entries] of this.state.batch.dependencyGraph) {
+      deps.set(key, [...entries])
+    }
+    return deps
+  }
```
rthreei: What call `getDependencies`?

Also, what is this function doing other than recreating the data structure into something slightly different?

src/batch/dependency-graph.ts
```diff
@@ -0,0 +1,161 @@
...
+  return prop in obj
+}
+
+/**
+ * Check if two values are deeply equal.
+ * For arrays, it needs to contain the same elements in the same order.
+ * For objects, it checks that all keys exist in both objects and that the
+ * values are equal.
+ */
+function isEqual(a: unknown, b: unknown): boolean {
```
rthreei: Does this function have complete test coverage?

src/batch/parse.ts
```diff
@@ -0,0 +1,35 @@
...
+    expressionsWithDynamic: new Set(),
+  }
+
+  for (const [name, expr] of expressionsMap) {
+    try {
+      parsed.expressions.set(name, engine.parse(expr))
+    } catch (error) {
+      if (error instanceof Error && error.message === 'invalid expression') {
+        throw new Error(`invalid expression with name ${name}`)
+      }
```
rthreei: Does the original error not have access to `name`?  Seems odd to recreate the error here.

src/batch/types.ts
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * A compiled batch of expressions with shared resources.
+ *
+ * All expressions in the batch share:
+ * - A single refs array (deduplicated across expressions)
+ * - A single consts array (deduplicated across expressions)
+ * - A single constSets cache (built lazily)
+ * - The same opNames map (operator string → opcode)
+ */
+export interface ParsedBatch {
```
rthreei: `type` instead?

src/batch/types.ts
```diff
@@ -0,0 +1,47 @@
...
+  /** Per-expression compiled data, keyed by expression name */
+  expressions: Map<string, Evaluable>
+  /** Dependency graph: context key → expressions that reference it */
+  dependencyGraph: DependencyGraph
+  /** List of expressions with dynamic references that should always be re-evaluated */
+  expressionsWithDynamic: Set<string>
+}
+
+export interface BatchEvaluatorOptions {
```
rthreei: `type` instead?

src/batch/types.ts
```diff
@@ -0,0 +1,47 @@
...
+  /** Map of expression name → raw expression input */
+  expressions: Record<string, ExpressionInput>
+  /** Optional parser options shared across all expressions */
+  options?: Partial<Options>
+}
+
+export interface BatchEvaluatorState {
```
rthreei: `type` instead?

src/batch/types.ts
```diff
@@ -0,0 +1,47 @@
...
+  /** The map of parsed expressions and their dependencies */
+  batch: ParsedBatch
+  /** Original expressions map — stored for addExpression/removeExpression */
+  originalExpressions: Map<string, ExpressionInput>
+  /** Last full context passed to evaluate() */
+  lastContext: Context | undefined
+  /** Cached results from the last evaluation */
+  cachedResults: Record<string, Result>
+  /** Expressions marked for evaluation in the next `evaluate()` call */
+  markedForEvaluation: Set<string>
```
rthreei: Seems like we are tracking quite a bit of state here.  Memory overhead maybe?  Anything here potentially redundant?

src/operand/reference.ts
rthreei: Should this just be a public property that can be accessed?  To avoid the function call overhead.

src/batch/parse.ts
atoledo: No, `name` is exclusive to the Batch design. The engine parses 1 expression at a time.

The issue here is that if you provide 1000 expressions and 1 of them is bad, receiving a simple "invalid expression" error doesn't help at all with troubleshooting.

src/batch/batch.ts
atoledo: Just providing access to the dependency map by the caller. Not actually used elsewhere.

It was on Luiz's original design and I liked it. Might be useful for some of our internal tooling.

With that said, I think I can probably return the raw `this.state.batch.dependencyGraph` now that I simplified the `DependencyGraph` to `Map<string, Set<string>>`. Before it had some Bytecode references

src/batch/dependency-graph.ts
atoledo: Missing 1 path, will add it

src/batch/parse.ts
```diff
@@ -0,0 +1,33 @@
...
+): ParsedBatch => {
+  const parsed: ParsedBatch = {
+    expressions: new Map(),
+    dependencyGraph: new Map(),
+    expressionsWithDynamic: new Set(),
+  }
+
+  for (const [name, expr] of expressionsMap) {
+    try {
+      parsed.expressions.set(name, engine.parse(expr))
```
rthreei: What is `name`?

Could this instead be of type `WeakMap<expr, enginer.parse(expr)`?  And no longer need `name`.

src/batch/parse.ts
rthreei: What is `name`?

The original error could serialize the invalid expression in its message?

src/batch/batch.ts
rthreei: > Might be useful for some of our internal tooling.

I would remove this function until there is an actual use case for it.

src/batch/parse.ts
atoledo: The expression name that the called provides in the Batch options

```
      const evaluator = new BatchEngine({
        expressions: {
          isAdult: ['>=', '$age', 18],
          isActive: ['bad-operator', '$status', 'active'],
        },
      })
```

So the names would be "isAdult" or "isActive" => `invalid expression with name isActive`

> The original error could serialize the invalid expression in its message?

Yes, but that doesn't seem a good idea. Some conditions have thousands of lines... and they can still look very similar.

src/batch/parse.ts
atoledo: Don't think this one makes sense to convert to WeakMap. We want to fully store all expressions at all times, so users can explicitly manage their whole domain (By adding or removing individual expressions, or disposing and starting over).

src/operand/reference.ts
atoledo: I tried that for the `references` and it didn't seem like it helped much, can try again

src/batch/parse.ts
rthreei: Hm, interesting, so that caller sets the name, which in our context would be the QuestionId or DatapointId.

src/batch/types.ts
atoledo: Don't think so.

* `originalExpressions` is needed for when new expressions are individually added into the system (or removed), which triggers recompilation and dependency graph calculation
* `lastContext` is needed so providing just what changed in the Context still works
* `cachedResults` is managed so we update only what was recalculated, still returning the whole status of the domain
* `markedForEvaluation` is the feature to track newly introduced expressions, which have to be evaluated on next evaluation. This is a bit of an edge case. See test `adds an expression and preserves existing cached results`

src/batch/types.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export type BatchEvaluatorState = {
+  /** The map of parsed expressions and their dependencies */
+  batch: ParsedBatch
+  /** Original expressions map — stored for addExpression/removeExpression */
+  originalExpressions: Map<string, ExpressionInput>
+  /** Last full context passed to evaluate() */
+  lastContext: Context | undefined
+  /** Cached results from the last evaluation */
+  cachedResults: Record<string, Result>
```
atoledo:
```suggestion
  cachedResults: Record<string, boolean>
```

Although Evaluable.evaluate returns Result, it is expected to be boolean since its the root level expression. Need to handle this like in the Engine.evaluate. It already is functionaly, but just to ensure the proper type is returned.

Here and in the return type of evaluate method

src/batch/batch.ts
```diff
@@ -0,0 +1,261 @@
...
+   *   Context is merged into stored context, all expressions run.
+   *
+   * Mode 2 — Subsequent evaluations with changed context: incremental evaluation.
+   *   Only expressions affected by updated context are evaluated. The caller can
+   *   choose to provide only the changed context. Keys which the value did not
+   *   change are ignored. Sending undefined for a key indicates that the key has
+   *   been removed from the context.
+   *
+   * @param ctx — Full evaluation context
+   * @param changedKeys — Optional list of keys that changed (trusted by caller)
```
atoledo: Need to remove all remaining `changedKeys` references.

src/batch/parse.ts
atoledo: Yes. Created a version without it for review https://github.com/briza-insurance/illogical/pull/328

src/batch/types.ts
rthreei: Could `originalExpressions` and `cachedResults` be tracked in one `expressionMap: WeakMap<ExpressionInput, Result>`?

src/batch/evaluate.ts
```diff
@@ -13,13 +14,19 @@ export function evaluateSingle(
...
   ctx: Context
-): Result {
+): boolean {
   const evaluable = batch.expressions.get(exprName)
   if (!evaluable) {
     throw new Error(`Expression '${exprName}' not found in batch`)
   }
 
-  return evaluable.evaluate(ctx)
+  const result = evaluable.evaluate(ctx)
```
rthreei: Do you know why the `Result` type is so broad?  And not just a `boolean` to begin with?

atoledo: That's a great question. I mean, I fixed it already just like the Engine does.

But I wonder if there is a reason for the internal Evaluable.evaluate to return Result.

atoledo: Ahh, the Operands don't return booleans, just expressions

src/batch/types.ts
atoledo: Conceptually yes, but I don't think so. We'd still need a way to iterate over every expression. WeakMap doesn't have iterable support.

